Issue #2, opened 2020-03-04
[Outreachy applications] Startup task: Train and test a classification model


dzeber:
This is a good way to get started with the environment and the problem domain. It will also provide the basis for a test case for future work. At a minimum, you should:

- load a dataset from the repo
- train a classification model from [scikit-learn](https://scikit-learn.org/stable/supervised_learning.html)
- compute an [evaluation metric](https://scikit-learn.org/stable/modules/model_evaluation.html) on a held-out test set

Feel free to include any additional steps you feel are relevant or you are interested in trying out, such as:

- basic exploratory analysis of the dataset
- data preprocessing
- hyperparameter tuning

When you start work on this task, please post a comment here indicating which dataset and model you will be working with so that other contributors can avoid duplicating your work.

KaairaGupta:
Hey! 
Can I work on this??

tab1tha:
I'll use the generated.csv dataset.

dzeber:
@KaairaGupta yes! in fact everyone is encouraged to work on this issue first

KaairaGupta:
Okay.

I'll use winequality.csv dataset

mhmohona:
Hello @dzeber! I want to work on vehicles.csv.

mhmohona:
Hello @dzeber! I want to work on defaults.csv.

dzeber:
Sounds good! To be clear, it's also fine to work on the same dataset as someone else, as there are only 5 datasets currently. The goal is just to collect a diversity of models and approaches across these datasets.

shreyagupta30:
Hello @dzeber, I'd like to work on eeg.csv.

Clare-Joyce:
Hello @dzeber, I will work with the winequality.csv

elie-wanko:
Hello @dzeber, I will be working on the **defaults.csv** dataset.

BBimie:
Hello @dzeber, I'll work with the winequality.csv

hammedb197:
Hello @dzeber, I will be working on vehicles.csv.

shiza16:
Hello @dzeber , I'd like to work on vehicle.csv .

terminator-4:
I would like to work on eeg.csv dataset.

alberginia:
I have been working on the generated.cvs dataset with SVMs.

Soniyanayak51:
I would like to work on defaults.csv @dzeber

silvererudite:
[...]

PR #12, opened 2020-03-06
issue #02 - Train and test on classification of vehicle dataset with Logistic Regression

shiza16:
The accuracy of my model is 82%, I used three classifier for it at first SVM, Logistic Regression and Decision Tree, but the result and accuracy with Logistic Regression was much better than other 2 classifiers, that's why then I used Logistic Regression Classifier for this model.


shiza16:
Thanks for you feedback, I will make the changes you told me. There is also
one thing I want to tell you I was working on different models.
At first attempt I use 3 different models Logistic Regression , Support
Vector Machine and Decision Tree and compared their accuracy. At first LR's
accuracy was higher than other but now after going more in deep and
changing some parameters of SVM, it's accuracy increased to 90%. So what
should I do know, can I change my classification model from logistic
regression to SVM.
And do I have to make a separate modules.py file in which all the functions
would be written or i can just do all my work in one python jupyter
notebook?.

Regards
Shiza Azam

dzeber:
> And do I have to make a separate modules.py file in which all the functions
> would be written or i can just do all my work in one python jupyter
> notebook?.

With this type of work, putting everything into modules becomes a bit contrived. I suggest moving everything between `Training Model` and `y_predictLR = LRclassifier.predict(X_test)` to a module function(s) that takes `vehicle` as input and returns `y_predictLR`, `X_*`, `y_*` as needed, so you can still display them in the notebook. The other cells are mainly visualization and interpretation.

> So what should I do know, can I change my classification model from logistic regression to SVM.

This type of question, tuning and comparing models, is getting right at the heart of applied ML and what this project is about: what is the best way to choose between models? It is great that you are thinking about this and have been working on tuning the models. Why not just show all of your work in this notebook? You can fit all three models, show the tuning that you did and how you made a decision based on accuracy. As we said, this project is much more about exploring this model selection process and much less about the final optimal model. As discussed above, please keep procedures like model fitting and dataset splitting in a module and use the notebook mainly for visualizations.

shiza16:
Thank you for your feedback. I will make following changes.

shiza16:
Hi @dzeber ,  I have made the following requested changes. Kindly review my PR and let me know if I am doing it right. 
- Trained **Logistic Regression** for classification  
- Hyper-parameters tuning of Logistic Regression
- Learning Curves
- Trained **Support Vector Machine** which has improved model by only 1%.
My model accuracy has totally changed, as i have changed the type of target labels from categorical to numerical data.

dev/shiza16/modules.py
```diff
@@ -0,0 +1,256 @@
+import pandas as pd
+import numpy as np
+import matplotlib.pyplot as plt
+import seaborn as sns
+from sklearn.model_selection import train_test_split
+from sklearn.svm import SVC
+<<<<<<< HEAD
```
dzeber: Looks like you have a merge conflict in this file. Git wants you to keep either from here to line 83 or from line 85 to line 255 but not both. It looks like you meant to keep the second part. To keep the second part, delete lines 7-84 and line 256.

dev/shiza16/modules.py
```diff
@@ -0,0 +1,256 @@
...
+    sns.heatmap(matrix, annot=True, linewidths=.5)
+    plt.show()
+
+def model_classification_report(y_test, y_predict):
+    """  Model Classification report for Precision , Recall and F1-Score """
+    print("\n DataSet Report: ")
+    print(classification_report(y_test, y_predict))
+    
+    
+=======
```
dzeber: Delete this line when resolving merge conflict.

dev/shiza16/modules.py
```diff
@@ -0,0 +1,256 @@
...
+    plt.xlabel(" Training Score ")
+    plt.ylabel(" Score ")
+
+    plt.plot(train_scores_mean, "o-", color="r", label="Training score")
+    plt.plot(test_scores_mean, "o-", color="b", label="Cross-validation score")
+
+    plt.legend(loc="best")
+
+    plt.show()
+>>>>>>> Issue#1--train-and-test
```
dzeber: Delete this line when resolving merge conflict.

dev/shiza16/modules.py
```diff
@@ -0,0 +1,256 @@
...
+        C=svm_grid.best_estimator_.C,
+        gamma=svm_grid.best_estimator_.gamma,
+    )
+
+    return classifier.fit(X, y)
+
+
+def LogisticRegression_train(X, y):
+
+    """ SVM Classifier"""
```
dzeber: Update docstring

dev/shiza16/modules.py
```diff
@@ -0,0 +1,256 @@
...
+    return classifier.fit(X, y)
+
+
+def Tuning_LogiticRegression(X, y, x):
+    """ Tuning Logistic Regression to increase the accuracy of model """
+
+    parameters = {
+        "penalty": ["l1", "l2"],
+        "C": [0.0001, 0.001, 0.01, 0.1, 1, 10, 100, 1000],
+        "solver": ["liblinear", "saga"],
```
dzeber: I don't think it's necessary to optimize over the solver.

dev/shiza16/modules.py
```diff
@@ -0,0 +1,256 @@
...
+        "class_weight": [
```
dzeber: Did you mean to include this? I don't think this does what you want because you have 4 classes, so this will exclude 2 of them IIUC.

dev/shiza16/Vehicle_Classifier.ipynb
```diff
@@ -0,0 +1,1355 @@
...
...xauZv3bulNmJ8X4xato2/zukSG+BXHnErLYdFvh/jwtr74eXmQlpsPgJe7kRev7UKjoNokZZu45X+/0KNpKH4uUW4rs5as4b2pNxEW6Mu4WYvp264ZkfXqFMc8NmZA8eslv+zg4OlEALq2bMTSZ22NFpm5JkY8s5AerR1/8LuyUVcP4pbrr+WpF1+t7lSqjMVqZdbybbw3oT9hft6Me28lfVtFEBnqXxxzKjWLRRv28+Gdg/HzqkVaju0479o0nKVTrgYgM6+AEW8uo0ezutVSjkuiDHiMmET+By+is9LwnDwL84Ht6OQzjnEenrj3GIYl5rDDbJ2WQP5/HnNiwlXDYrUy68v1vDdlNGEBtRk393P6tm1KZN3g4pjHru9b/HrJr7s4eCbZYRvzfvydzs3qOy3nqhA9sBeNmzakf7eRdOjclhfmTuP6Ibc7xOTm5DGi39ji6e/XfsqqH20X1KbNmMq3X/zAN1/8QI8+XXn02ft59N5nnVqGi2WxWpn1zWbeu/sa2/f3m9/QN6oxkeEl39+PjexZ/HrJxn0cjE0pnr49uj35RWa+2nLAqXn/XRarZva6A7x7XWfCansybsnv9G0aQmRwbYe4IS3CebLfFQ7zujYI4ovxPQDIzC/i2g82cmWjYFyRxWpl5icrWPDIeMKC/Ljlhf8S3aElkfVDimPeX76RIV2jGNOvC8dik7nvzc9YMffBasz60lisVmZ9tYH3Jl9rO6+9/iV92zQhMjyoOOax0SUN70s27Ck+r4X4+bB46vV4uBnJKyjk+tmf07dNE0L9fZxejoumFB79xlLwzZvonHQ8x07DcnwPOi3eIcx8eHulDcvuPa7FeuaIM7KtUvX7t8evSTjf9X6EOp0i6T5rAitGTC8Xd2bNTg59sIZRmxzrbm0fHMnJ5Vs5/NFa/JvXY8DHj/HNlQ85KfsLZDBQ/4V7ODH+WYoSUmm27HWy1myl4Ojp4pCgMYOxZOZwKPpu/Ef0oe6TE4i5bw4AhacSOHJ1+f/n7LV/kLr4B1quX+C0olwyg4HAxx4k+f7HsCQlE/rhu5g2/ob5REljoiUxkfQXX6H2OMeG9IIdu0i69S4AlJ8vdb/6mIKt252a/t+iDHjeNIW8t59CZ6Tg/cRbmPdstTUu2xVtW0/Rxp8AMLbtTq3r78Q071ncOvUBN3fyXr4X3Gvh89wCiratR6
```
dzeber: Since you plan to do cross-validation, don't compute test-set accuracy at this point - use the CV score instead.

dzeber: It looks like this is running on the full dataset? It should run only on the training set.

dzeber: I don't think this necessarily means it's overfitted - the scores are very close.

shiza16:
@dzeber , I have made the requested changes, kindly take a look.

mlopatka:
@dzeber I didn't realize that outstanding reviews block merging.
@shiza16 I'm going to merge this in for you so I dont have to dismiss @dzeber's review.

PR #14, opened 2020-03-07
WIP: For #2: on winequality.csv

KaairaGupta:
This PR contains `winequality.csv` trained on several different models defined in sklearn.
I have made changes in 1 file which is .gitignore, to ignore the cache generated by vscode

Also, I have made a directory by my name in `dev` as suggested by the README.md, which contains 4 files.
- One is a .ipynb file wherein I visualised data and found that it was skewed.
- In the data loading file, I preprocessed the data as well so that the skewness can be taken care of.
- 3rd file is a file which defines all the functions. Since I found the data to be a bit unbalanced as well (3:1 ratio of false to positive) I used recall and precision as metrics as well.
- main.py() calls all the functions.


KaairaGupta:
Closing this because I found a better way to do this task. @dzeber

Issue #7, opened 2020-03-04
[Outreachy applications] Visualization for misclassifications


dzeber:
Misclassifications can reveal a lot about the boundaries of performance of a classifier. Develop a visualization that helps dig into misclassified datapoints in the test set. A simple approach for a binary classifier would be to plot a histogram of the predicted class probabilities across the misclassified test samples in each class.

KaairaGupta:
This issue seems interesting. May I be assigned this?

dzeber:
You're welcome to work on it and submit a contribution! We're not assigning issues currently to allow people to bring multiple perspectives or approaches to these questions.

shristi428:
how can we contribute in the project should we PR then you assign us task ?

dzeber:
@shristi428 You can submit a PR referencing the issue you worked on.

namrathagopalabhatla:
Hey @dzeber! What dataset do we use to test out our code?

dzeber:
@namrathagopalabhatla whichever you like. It might be a good start to use the same data & model you used for #2.

PR #20, opened 2020-03-08
Added Various Classifiers for all datasets. 

abbasidaniyal:
Monitored accuracy and F1 score. Plotted confusion matrix. Also used xgboost Classifier as it is very popular among Kagglers due to efficient performance.
In reference with #2

PR #21, opened 2020-03-08
Added confusion_matrix plotting helper function in the helpers directory

abbasidaniyal:
Added a helper function to plot confusion matrix which visualizes the misclassified examples. This is a useful approach to check misclassifications and check for overfitting in data with high bias.
In reference with #7

### DIFF
--- a/environment.yml
+++ b/environment.yml
@@ -1,15 +1,20 @@
 name: presc
 channels:
   - conda-forge
-dependencies:
-  - numpy=1.18.1
-  - pandas=1.0.1
-  - python=3.7.6
-  - scikit-learn=0.21.3
-  - scipy=1.4.1
-  - jupyter=1.0.0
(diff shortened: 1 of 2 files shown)


dzeber:
(Example output is given in the notebooks in #20 .)

While the confusion matrix is a good initial technique to look into classification accuracy, it is a widely used tool. For #7 , I'm interested in developing something to dig deeper into the individual test samples that got misclassified. Why did those points in particular get misclassified? Are they all "unusual" examples, or are they evenly spread across the feature space? Were they close to the decision boundary?

Please keep your confusion matrix code in your submission to #20. I am closing this PR as this code is included there, but I encourage you submit a new one with an approach that looks into some of the questions described above.

abbasidaniyal:
Will do! Thank you

PR #22, opened 2020-03-08
Outreachy startup task for defaults.csv

Soniyanayak51:
PR for issue #2 
1. Added module with all functions to import from jupyter notebook.
2. Added Jupyter Notebook with outputs.

I have used the following 2 models on the defaults.csv dataset:
1. Logistic Regression
2. SVM
Both using scikit-learn.


dev/Soniyanayak51/classificationModelFunctions.py
```diff
@@ -0,0 +1,175 @@
...
+from sklearn.svm import SVC
+from sklearn.linear_model import LogisticRegression
+from sklearn.datasets import load_iris
+
+
+def loadDataAndPrintHead():
+    '''
+    Loads data using pandas read_csv.
+    '''
+    df = pd.read_csv("/home/soniya51/Downloads/default of credit card clients.csv")
```
mlopatka: please split out the local filepath as a variable to make local review easier.

Soniyanayak51:
> This is a good start on the default dataset. A bit more interpretation of the results would be helpful.
> 
> * confusion_matrix` results are output, but not discussed.
> * data is cleaned to consolidate categories, have alternate strategies been considered?
> * what is the motivation for the two classifiers chosen `svm` and `logistic regression`?
>   Please extend the analysis to discuss the problem and interpret the results achieved.

dev/Soniyanayak51/classificationModelFunctions.py
Soniyanayak51: @mlopatka Thanks for the review! 
Sure, I'll incorporate the suggested changes.

Soniyanayak51:
@mlopatka Thanks for the review!
Sure, I'll incorporate the suggested changes.

dev/Soniyanayak51/classificationModelFunctions.py
dzeber: Also, please make sure to use the CSV file included in the repo. That way anyone who runs the code should get consistent results. This will require you change other things in the notebook that depend on the CSV structure, eg column headings.

dev/Soniyanayak51/classificationModelFunctions.py
```diff
@@ -0,0 +1,175 @@
+import pandas as pd
+import numpy as np
+import matplotlib.pyplot as plt
+import tensorflow as tf
```
dzeber: Please drop this import

dev/Soniyanayak51/OutreachyStartupTaskSoniyanayak51.ipynb
```diff
@@ -0,0 +1,983 @@
...
+   "cell_type": "code",
+   "execution_count": 1,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "import classificationModelFunctions # all functions are defined in this module\n",
+    "import pandas as pd\n",
+    "import numpy as np\n",
+    "import matplotlib.pyplot as plt\n",
+    "import tensorflow as tf\n",
```
dzeber: Please drop this import

dev/Soniyanayak51/OutreachyStartupTaskSoniyanayak51.ipynb
```diff
@@ -0,0 +1,983 @@
...
+       "[PhysicalDevice(name='/physical_device:GPU:0', device_type='GPU')]"
+      ]
+     },
+     "execution_count": 2,
+     "metadata": {},
+     "output_type": "execute_result"
+    }
+   ],
+   "source": [
+    "tf.config.experimental.list_physical_devices('GPU') # check for GPU"
```
dzeber: Please drop this cell

dev/Soniyanayak51/classificationModelFunctions.py
```diff
@@ -0,0 +1,175 @@
...
+    Args: x - data to be standardize
+          x_train - trainind data
+    Returns: standardized Data x
+    '''
+    x = x.astype(float)
+    x_train = x_train.astype(float)
+    for column in x.columns:
+        mean = np.mean(x_train[column])
+        std = np.std(x_train[column])
+        x[column] = (x[column] - mean) / std
```
dzeber: I believe this standardization can be done more easily using `StandardScaler`, which I see you imported above. I'd recommend using that unless there is a reason why it didn't work for you.

dev/Soniyanayak51/OutreachyStartupTaskSoniyanayak51.ipynb
```diff
@@ -0,0 +1,983 @@
...
+    "from sklearn.utils import shuffle\n",
+    "from sklearn.metrics import confusion_matrix\n",
+    "import seaborn as sns\n",
+    "import matplotlib.gridspec as gridspec\n",
+    "from sklearn.preprocessing import StandardScaler\n",
+    "from sklearn.manifold import TSNE\n",
+    "from pandas import set_option\n",
+    "from sklearn.svm import SVC\n",
+    "from sklearn.linear_model import LogisticRegression\n",
+    "from sklearn.datasets import load_iris"
```
dzeber: Please drop any imports that are not explicitly used in the notebook.

Soniyanayak51:
@dzeber thanks for the review! I'll update the PR with the suggested changes.

mlopatka:
@Soniyanayak51 Please let us know if you intend to continue with this PR. I see that some of the outstanding review comments have not yet been addressed. Please lag me with an `@` mention when you are ready for another pass of this PR.

Soniyanayak51:
@mlopatka I was working on some other PR's but I have made the requested changes regarding standardization, imports and discussion. I have used the UCI default of credit card clients Data Set but have changed the xls file to csv. Kindly review and let me know if I need to elaborate more on this notebook. I'll be happy to make the changes!

Soniyanayak51:
@mlopatka @dzeber please review.

dzeber:
> Also, please make sure to use the CSV file included in the repo. That way anyone who runs the code should get consistent results. This will require you change other things in the notebook that depend on the CSV structure, eg column headings.

You need to use the [CSV file](https://github.com/mozilla/PRESC/blob/master/datasets/defaults.csv) that is actually included in the repo. It has undergone minor reformatting and is slightly different from the raw UCI dataset, so it will require you to adjust some of your data processing code. Sorry if that wasn't clear!

Soniyanayak51:
@mlopatka @dzeber I have added the dataset included in the repo and python black formatting. Also added axes lables and titles to graphs. Please review.

mlopatka:
`"../../.././datasets/defaults.csv"`
evaluates to the same as
"../../../datasets/defaults.csv"
Please take the time to understand the changes and test locally before pushing new commits.

Issue #6, opened 2020-03-04
[Outreachy applications] Visualization of an evaluation metric


dzeber:
A common theme in classifier tuning and evaluation is to plot a metric against the values of some parameter with repeated runs to assess variability. We would like to have a general utility for producing such plots.

It should take as input at table with columns (x, y1, y2, ..., yk), ie multiple y values for each x, and plot the average y value vs x with the spread of the y-values represented in some way, eg. a band.

Soniyanayak51:
@dzeber I would like to work on this.

Sidrah-Madiha:
@dzeber  I have a question for this issue: I am not sure what is meant by spread of y-values, does this mean there should be a plot of x and average y, and there should be a color shading from minimum value of y to maximum value of y at each point?

simran0117:
I have tried #2 with SVM and KNN on Wine Quality Dataset, I would like to plot the outputs of the same and also work on this. Should I go for something specific, before a general approach @dzeber?

dzeber:
> plot of x and average y, and there should be a color shading from minimum value of y to maximum value of y at each point?

@Sidrah-Madiha yes, something like that would work. It's nice to see some representation of the spread to see how far the individual values tend to be from the mean.

dzeber:
@simran0117 The ultimate goal would be a general approach. However, I recommend using your previous notebook as a place to develop your idea for this, and then move your code a separate module and make sure it does not depend on the specific model or metric used in the notebook. Also, you should include the notebook in your PR showing an example of your function in action!

mlopatka:
There have been several questions about this issue, so I'll try to add a bit of context here:

The intention with this issue is to get everyone thinking about the stability of reported performance in ML. Often times a model will be decided upon, trained, and then documented. In the worst case only a single percentage performance is reported, i.e. "This model exhibits correct classification 92% on out of box samples". This does not tell the whole story. Issue #6 is an opportunity to proposed strategies for examining how much variability there is for a whole system: available training data can be split in a multitude of different ways, classifier parameters may be tuned to produce different performance for different data splitting strategies, parameters in both model training and data splitting may affect not only the raw performance (accuracy, recall, precision, F1, etc..). Any changes that lead to different `y` label predictions over a data set `x`. As mentioned in the issue, explore what the variation in those `y` values is/can be depending on the holistic set of parameters that go into building an ML pipeline.

PR #24, opened 2020-03-08
Plot analysis for visualising evaluation metrics

Soniyanayak51:
PR for issue #6 

1. Simple scatter plot for x and y values
2. Violin and Box plots
A boxplot is a standardized way of displaying the dataset based on a five-number summary: the minimum, the maximum, the sample median, and the first and third quartiles. I have changed the median value of box plot to mean value(average of y).
Violin plot is similar but it also presents the probability density.

@dzeber please let me know if the approach is on the right track!


Soniyanayak51:
@dzeber I would like to try dot/strip plots or python representation of R’s ggplot2 for this if you could review this once

dev/Soniyanayak51/issue6/evalMetric.ipynb
```diff
@@ -0,0 +1,130 @@
...
+    "from sklearn.utils import shuffle\n",
+    "from sklearn.metrics import confusion_matrix\n",
+    "import seaborn as sns\n",
+    "import matplotlib.gridspec as gridspec\n",
+    "from sklearn.preprocessing import StandardScaler\n",
+    "from sklearn.manifold import TSNE\n",
+    "from pandas import set_option\n",
+    "from sklearn.svm import SVC\n",
+    "from sklearn.linear_model import LogisticRegression\n",
+    "from sklearn.datasets import load_iris"
```
dzeber: Please only import things you use

dev/Soniyanayak51/issue6/moduleForEvalMetric.py
```diff
@@ -0,0 +1,73 @@
...
+from sklearn.utils import shuffle
+from sklearn.metrics import confusion_matrix
+import seaborn as sns
+import matplotlib.gridspec as gridspec
+from sklearn.preprocessing import StandardScaler
+from sklearn.manifold import TSNE
+from pandas import set_option
+from sklearn.svm import SVC
+from sklearn.linear_model import LogisticRegression
+from sklearn.datasets import load_iris
```
dzeber: Please only import things you use

Soniyanayak51:
@dzeber thanks for the great interesting suggestions. Will implement them

dev/Soniyanayak51/issue6/moduleForEvalMetric.py
Soniyanayak51: Will keep this in mind

Soniyanayak51:
@dzeber @mlopatka I have generated a small dataset by changing C values and generating accuracy for each fold using repeated K-fold. I have added a new plot to visualize variance.
Plots are plotted with the above data.
Please review!

mlopatka:
@Soniyanayak51 It appears that you have multiple PRs addressing some of the same changes to your work. Can you please follow up with comments in each of your current outstanding PRs as to which of them are active work and which of them are stale.

It would be very helpful to know where we should focus our attention to ensure that reviews are on the most recent work. Feel free to close out stale PRs that you submitted so that we don't review them.

Soniyanayak51:
@mlopatka I have been using the same notebook as my startup task to generate the data here. I have used the same notebook and the module from PR #22 in most of my work as suggested by @dzeber. 
Do I need to remove it and make new modules?
Also could you please review the latest changes I made in this PR? I have generated a small dataset by changing C values and generating accuracy for each fold using repeated K-fold. Thank you:)

dzeber:
@Soniyanayak51 using the same notebook and module as #22 is fine. When these eventually get merged, they should appear as incremental changes to the same files. 

Same comment as in #22, please use the CSV file included in the repo.

I like that you are trying out different approaches to the visualization, as they are more or less appropriate in different situations. One suggestion I would make for all of them is to plot the x-values at equally spaced intervals. Using the original numerical scale causes those uneven gaps which is difficult to read. The log-scale for the mean-variance plot helps but it is still difficult to read because the log is base 10 rather than 2. I think the scale of the differences between the parameter values on the x-axis is not so important to show in these plots, as the goal is to select whichever one has the optimal metric value.

As the goal of the task is to create a function that can be applied to a general CV dataset and metric, the next step would be to settle on a final visualization approach based on what you've tried and develop it into a generalized function. Note that the box/violin and mean/variance plots are good ways to summarize when you have many y values for each x, but can be misleading when the dataset is so small (in which case the simple plot may be more representative).

PR #16, opened 2020-03-07
Performance Comparision of Trained models on eeg.csv

Addi-11:
* Various classification models have been trained on the dataset.
* Accuracy of each model measured.
* The results can be seen in [demo.ipynb]()

Further various evaluation metrics to be added and plotted for various data splits for `issue 4`.

PR #25, opened 2020-03-08
fixes #4

### LINKED ISSUE #4: [Outreachy applications] Traversal of the space of cross-validation folds
Similarly to #3, we want to investigate how much the performance score computed using cross-validation depends on the number of folds. Eg. how would our performance estimate change if we used 10-fold rather than 5-fold?

Write a function that takes a scikit-learn estimator and a dataset, and computes an evaluation metric using repeated K-fold cross-validation over a grid of K values from 1 to n. It should output a table of K with the average metric value across the folds, one for each repeat.

Addi-11:
The PR is in respect to #4 
* Tabulated evaluation metrics: 
  * Accuracy
  * Precision
  * Recall
  * F1_score , against values of K ranging from 2 to 80 (a huge set of k was computationally expensive to compute)


dzeber:
This PR gives a nice approach to the task, but needs a little more work.

- The outer function should take a model and evaluation metric as args, and run the computation for those.
- Please use `RepeatedKFold` CV. This will give a table with multiple rows for each K value.
- Is there a way to better handle computation costs for larger K? Eg. instead of fitting the model to all possible splits, maybe sampling a subset of splits for a given K will be sufficient.

Also, please make sure to follow the comments in #16. Once that is merged, this PR can break off the changes on top of that.

Addi-11:
Thanks for the review @dzeber.
 I understand your point regarding the computational costs, I will improve upon it and rectify the code as soon as possible.

mlopatka:
@Addi-11 please let us know if you intend to continue with this PR or if this work has been integrated elsewhere?

Addi-11:
I am sorry for the delay, but currently I am stuck with an error, and am trying to resolve it.

This is the error I get --
Screenshot from 2020-03-17 11-20-33

The 2nd n 3rd line of the loop, cause the error (assigning), print works just fine. 
```
for train_index, test_index in rkf.split(X):
...     print("TRAIN:", train_index, "TEST:", test_index)
...     X_train, X_test = X[train_index], X[test_index]
...     y_train, y_test = y[train_index], y[test_index]
```

Could you guide me, where am I wrong ?

Addi-11:
I am not able to figure out the error. The code works on kaggle notebooks, but doesn't locally, the indexing doesn't work.

dzeber:
@Addi-11 is there a reason you're not using `X.iloc[train_index]`, as you are currently on line 21? That should work. I'm not sure what code you're using as an example, but maybe they're using numpy arrays rather than pandas dataframes.

Addi-11:
I was following the documentation sample code of `Repeated KFold`, by scikit learn and it worked fine in many examples. It even worked in kaggle notebooks i downloaded.
Moreover using `X.iloc` will require a loop, but the function works in vectorized format which is much more efficient way to follow and  using `X.iloc` gives an error.

Addi-11:
I am sorry for the delayed solution for the PR. I have had problems with the implementation, so i changed my complete approch to the solution. 
Also, using repeated cv gave me a few complications, so I created a substitue function for that, it maps the model's performance for different number of repetitions in the cross validation set.
Instead of giving a table as output, I have just used graphs to visualise the result for different models.

Please do let me know for any shortcomings.

dev/Addi-11/kfold.py
```diff
@@ -0,0 +1,102 @@
...
+
+    for i, fold in enumerate(folds):
+        fit_times[i] = np.mean(
+            cross_validate(clf, x, y, cv=fold, return_train_score=True)["fit_time"]
+        )
+        train_scores[i] = np.mean(
+            cross_validate(clf, x, y, cv=fold, return_train_score=True)["train_score"]
+        )
+        test_scores[i] = np.mean(
+            cross_validate(clf, x, y, cv=fold, return_train_score=True)["test_score"]
```
dzeber, line 36: It looks like you are rerunning the cross-validation 3 times?

Addi-11: I apologise for the silly mistake.  Yes, I was suppose to run it once. 
Will work on it right away and also the changes suggested on the graphs.

PR #27, opened 2020-03-08
for #6 : Visualise evaluation metric

KaairaGupta:
For #6 
The task was to plot the mean and spread of `y` vs `x`.
I have plotted the following vs x with input as defined in the task:
- mean of y 
- maximum and minimum values of y
- spread of y around mean (standard-deviation)


dev/KaairaGupta/evaluation metric/example_visualise_evaluation_metric.ipynb
```diff
@@ -0,0 +1,113 @@
...
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "data = []\n",
+    "\n",
+    "x = []\n",
+    "\n",
+    "# Creating simple data for [x1,x2,x3,...x4]\n",
+    "for i in range(10):\n",
+    "    x.append(i)\n",
```
dzeber: Alternatively, `np.arange(10)`

KaairaGupta:
@dzeber Do you want me to remove the standard deviation band and just show a band between min and max y?
Also, I was thinking to plot multiple graphs, one as above and other with violin plots. Do you think I should add that?

mlopatka:
@KaairaGupta I'll jump in for @dzeber here since he is away until Wednesday.

> @dzeber Do you want me to remove the standard deviation band and just show a band between min and max y?
Yes, you should update the visualization to include only the min/max bounds.

> Also, I was thinking to plot multiple graphs, one as above and other with violin plots. Do you think I should add that?
Yes! I think violin plots would be very informative in complimenting the above-mentioned plot.

KaairaGupta:
@dzeber @mlopatka 
I have made all required changes. PTAL.

PR #28, opened 2020-03-08
KNN model trained and tested on generated.csv dataset

tab1tha:
This pull request addresses https://github.com/mozilla/PRESC/issues/2
The generated.csv dataset was explored and split into two sets; one set for fitting a machine learning model and the other set to be used for testing it.

The  target variable of the generated.csv dataset is called "label" and a demarcation was made between it and the predictor variables before training begun.

The k-nearest neighbor algorithm was used as the model and its performance was quantified.  Finally, a model complexity curve was drawn to ease visualization of the KNN model on the generated.csv dataset.


tab1tha:
I had a little difficulty doing visualizations with seaborn when carrying out exploratory data analysis. This may be because the column names are numeric and not strings. Can I change the column names to feature0, feature1, and so on?

mlopatka:
> I had a little difficulty doing visualizations with seaborn when carrying out exploratory data analysis. This may be because the column names are numeric and not strings. Can I change the column names to feature0, feature1, and so on?

^ absolutely, if this will help you, feel free to rename columns.

dev/tab1tha/k_nn.py
```diff
@@ -0,0 +1,53 @@
+from sklearn.neighbors import KNeighborsClassifier
+import matplotlib.pyplot as plt
+import numpy as np
+
+def k_nearest(data_train, target_train, data_test, target_test):
+    #training model using the k nearest neighbor algorithm
```
mlopatka: Please replace multi-line comments with docstrings as per relevant [PEP conventions](https://www.python.org/dev/peps/pep-0257/)

dev/tab1tha/k_nn.py
```diff
@@ -0,0 +1,53 @@
...
+    #note that 'neighbor' must be spelt without including a 'u' in order for it to work
+    '''This function trains the model then uses it to predict target values of the test data. In it, the accuracy
```
mlopatka: This should be consolidated in a unified docstring as mentioned int he comment above.

dev/tab1tha/k_nn.py
```diff
@@ -0,0 +1,53 @@
...
+    of the K-nearest neighbor algorithm on this data set is computed and returned '''
+    knn = KNeighborsClassifier(n_neighbors = 5)
```
mlopatka: What is the motivation for specifying exactly 5 neighbours?
Does the performance change with this parameter?

Please explain in comments, or provide a sub-analysis motivating this decision.

dev/tab1tha/k_nn.py
```diff
@@ -0,0 +1,53 @@
...
+    y = target_train 
+    X = data_train
+
+    #Use data to train the model
+    knn.fit(X,y)
```
mlopatka: nit: why cast these variable into shorter, less descriptive, names. You can also pass them directly to knn like so:
`knn.fit(data_train, target_train)`

dev/tab1tha/k_nn.py
mlopatka: Ah! I see that you address this below in `visual_compare`.
Please indicate that `5` is only the default value in a comment.

dev/tab1tha/k_nn.py
```diff
@@ -0,0 +1,53 @@
...
+    #compute accuracy and return it
+    return knn.score(data_test, target_test)
+
+def visual_compare(data_train, target_train, data_test, target_test):
+    '''This function outputs a graph of the predictions gotten from the K-nn algorithm. In that same graph, the 
+    actual values are plotted. The variation between the predicted values and actual values is clearly visualised such
+    that the user can choose the optimal value of the n_neighbors parameter of KNeighborsClassifier that will yied maximum
+    prediction accuracy'''
+    # Setup arrays to store train and test accuracies
+    neighbors = np.arange(1, 9)
```
mlopatka: nit: Did you experiment with values beyond the 1:9 range?
Please specify the motivation of this range in particular.

dev/tab1tha/main.py
```diff
@@ -0,0 +1,31 @@
+import pandas as pd
+import sklearn
+import seaborn as sns
+import k_nn
+from sklearn.model_selection import train_test_split
+
+#import data into a pandas dataframe
+df = pd.read_csv('..\..\datasets\generated.csv')
```
mlopatka: nit: Please break out the local filepath into a variable and then pass that variable. This will make local testing easier for reviewers.

dev/tab1tha/main.py
```diff
@@ -0,0 +1,31 @@
...
+print('*'*50+'INFO'+'*'*50)
+print(df.info())
+print('*'*100)
+
+#todo visualisation
+
+data = df.drop('label', axis = 1)
+target = df['label']
+#split data set into train and test sets 
+data_train, data_test, target_train, target_test = train_test_split(data,target, test_size = 0.30, random_state = 10, stratify = target)
```
mlopatka: What is the motivation for a specific split ratio of exactly `test_size = 0.30`?
Does performance vary when this ratio changes?
Please motivate the choice of this parameter or provide a sub-analysis that explores the effects of this parameter.

dev/tab1tha/k_nn.py
tab1tha: In this case, due to the fact that turning the #comment into a docstring might make it to be interpreted as part of the  function's docstring, can I move the second comment line that begins with "note that ..." from line 7 to line 11, so that there would not be concurrent #comments as observed now in lines 6 and 7?

tab1tha: I have merged both comments into the main docstring as you recommended.

dzeber:
Thank you for this nice, concise PR. The only thing I want to add to Martin's comments is that if you plan to use `visual_compare` to select an optimal model, you should really compute "test accuracy" on a validation set or use cross-validation, and keep the actual test set until the optimal model has been selected.

dev/tab1tha/k_nn.py
tab1tha: done

dev/tab1tha/k_nn.py
tab1tha: done

dev/tab1tha/k_nn.py
tab1tha: Increasing the range only increased the scope, such that if there is another appropriate point, later in the number line, the user will be able to see that too. These values were however chosen randomly, respecting the requirements that values of n_neighbors must be greater than zero and be of integer type. More is done in the new commit, with the help of hyperparameter tuning.

dev/tab1tha/main.py
tab1tha: This has been solved using the pathlib builtin module

dev/tab1tha/main.py
tab1tha: The default value is 0.25, I chose 0.30 as a close approximate. I have investigated the effect of test size on performance in #43.

tab1tha:
Thank you very much for the detailed review. In the processes of implementing the requested changes I took some time out to learn about cross validation, hyperparameter tuning, and pair plots. I have implemented your changes in the commit d2336c5 to the best of my present knowledge. I look forward to your next review. Thank you again for this opportunity, I've learnt a lot in the past few days.

dev/tab1tha/load_dataset.py
```diff
@@ -0,0 +1,11 @@
+import pandas as pd
+from pathlib import Path
+
+
+def load_dataset(filename):
+    """Function that reads the dataset file into a dataframe irrespective of
+    the operating system. It returns a pandas dataframe. """
+    datasets = Path("../../datasets/")
```
mlopatka: nit: if you are writing a general purpose helper function to facilitate filesystem data import, then all locally varying pieces should be parameters. Here, the path string should be passed in as well.

tab1tha:
@mlopatka please can you merge this? I do not have merging rights.

PR #32, opened 2020-03-09
WIP: Visualization for misclassifications using Seaborn countplots

Soniyanayak51:
PR for issue #7 
@dzeber  I have currently used Seaborn countplots to see the distribution of 1's and 0's for the predictions and y_test.
Plots for both logistic regression and svm models from the startup task for defaults.csv have been shown.

I will continue to work on this.


dev/Soniyanayak51/misclassificationVisualization/misclassificationVisualisationFunctions.py
```diff
@@ -0,0 +1,51 @@
...
+    Agrs: The label y
+    '''
+    predsSeries = pd.Series(preds)
+    predsSeries.reindex(index=y_test.index)
+    df_miss = pd.concat([y_test, predsSeries], axis=1, join='inner')
+    df_miss = df_miss.rename(columns={"default payment next month": "y_test", 0: "predsSeries"})
+    ax1 = sns.countplot(x="predsSeries", hue="y_test", data=df_miss, palette="Blues")
+ #   plt.show()
+    
+def plotGrapthForAttributes(df):
```
dzeber: I believe this is already included in your other module, in which case you can drop it from here.

dev/Soniyanayak51/misclassificationVisualization/misclassificationVisualisationFunctions.py
```diff
@@ -0,0 +1,51 @@
...
+    
+def misclassificationGraphs(y_test, preds):
+    '''
+    Plots a graph to show default values
+    Agrs: The label y
+    '''
+    predsSeries = pd.Series(preds)
+    predsSeries.reindex(index=y_test.index)
+    df_miss = pd.concat([y_test, predsSeries], axis=1, join='inner')
+    df_miss = df_miss.rename(columns={"default payment next month": "y_test", 0: "predsSeries"})
```
dzeber: This should really be handled outside of this function, so that it doesn't depend on the particular dataset.

Soniyanayak51:
@dzeber thanks for the review. Working on it.

dev/Soniyanayak51/misclassificationVisualization/misclassificationVisualisationFunctions.py
Soniyanayak51: Removed this.

dev/Soniyanayak51/misclassificationVisualization/misclassificationVisualisationFunctions.py
Soniyanayak51: Refactored the code.

Soniyanayak51:
@dzeber @mlopatka I have made some changes please review.

PR #37, opened 2020-03-10
For #2 : Logistic Regresssion on winequaliy.csv 

SanchiMittal:
I have used Logistic Regression classifier to perform binary classification on winequality.csv and classify the test data into recommended or not recommended wine.
@dzeber and @mlopatka Please let me know what improvements I need to make.
Thanks


SanchiMittal:
I will add other ML models like KNN, SVM, Naive Bayes etc. in the next step and present a comparative study. @dzeber @mlopatka Kindly provide the feedback on my work till now, guiding if I need some changes in my workflow.
Regards.

dev/SanchiMittal/dataloader.py
```diff
@@ -0,0 +1,107 @@
...
+
+    sns.set(style="ticks", color_codes=True)
+    sns.pairplot(
+        d, vars=features, hue="recommend",
+    )
+
+    # Correelation Matrix
+    corr = d.corr()
+    print("Correlation Matrix:")
+    print(corr)
```
dzeber: I would not both printing this since you are also displaying the visualizations.

SanchiMittal:
Thank you for the review. I will implement the above suggestios in my work. Also, I would like to dig more into modelling and study the performance of different models.

PR #39, opened 2020-03-10
Lift-Gain Charts for classification models

Addi-11:
Among various methods to evaluate the performance of the trained models, I studied that Lift and Gain Charts are a useful way of visualizing how good a predictive model. They measure how much better one can expect to do with the predictive model comparing without a model. 

It can be helpful to find the best predictive model among multiple challenger models, thus I sought to create a function to visualise this.
Screenshot from 2020-03-10 17-03-57


dev/Addi-11/classifiers.py
```diff
@@ -0,0 +1,60 @@
...
+    
+    def Gaussian(self, X_train,y_train):
+        classifier = GaussianNB()
+        classifier.fit(X_train, y_train)
+        return classifier
+
+    """ This function is to evaluate classifier's performancen"""
+    def evaluation(self, classifier, X_test, y_test):
+        accuracy, precision, recall, f_score , y_score = evaluate(classifier, X_test, y_test)
+        print("Accuracy : ",accuracy)
```
mlopatka: nit: space following comma.

dev/Addi-11/classifiers.py
```diff
@@ -0,0 +1,60 @@
...
+        print("Precision: ", precision)
+        print("Recall: ", recall)
+        print("F1 score : ",f_score)
```
mlopatka: nit: space following comma.

Addi-11:
> Nice, I like the idea of Lift-Gain charts for model evaluation!
> The work on `data_split_examine()` is also great, this is an interesting direction for understanding artifacts of the processes we employ when testing as they relate to the expected generalized performance.
> 
> One interesting direction that you may want to explore as you dig deeper into performance calibration plots is existing work on [Tippett plots](https://github.com/davidavdav/ROC/blob/master/R/tippet.plot.R) I don't know of a good python implementation, but the R code should be easy to port.

I am currently working on ROC curves. I had opened a PR #44 wrt to this, and have had some issues with the implementation, when fixed I will inform. Thanks ofr the source.

> I see this is a WIP, so I'll hold off on deeper review, but this is looking good. A couple of general comments regarding style and formatting:
> 
> * Please be mindful of whitepace for code readability. Even code in jupyter notebook cells should adhere to [black formatting guidelines](https://pypi.org/project/black/). This helps with readability and expedites review turnaround time.
> * Capitalization of variable names matters for different IDEs, this can be distracting when you collaborate with folks using different environments.

I was following the documentation of scikit learn so I chose to use `X`. I will correct it too.

> * Your notebooks could make better (more verbose) use of the text cells. We are interested in your interpretation as well as the results. Take the opportunity to explain your thinking and tell us what the findings mean in the context of the problem space.

All changes noted, will work on it ASAP. Thankyou for the review

Addi-11:
I have implemented the requested changes.  Instead of interpreting the result for each classification model, i have stated the way to interpret the graph and the collective conclusion.<br>
Please look for any short-comings.

Addi-11:
Thankyou for the review.<br>
 I will keep in mind the PR formats for future.

Issue #3, opened 2020-03-04
[Outreachy applications] Traversal of the space of train/test splits


dzeber:
Given a classification model, we want to investigate how much the performance score computed on the test set depends on the choice of train/test split proportion. Eg. how would our performance estimate change if we used a 60/40 split rather than 80/20?

Write a function that takes a scikit-learn estimator and a dataset, and computes an evaluation metric over a grid of train/test split proportions from 0 to 100%. To assess variability, for each split proportion it should resplit and recompute the metric multiple times. It should output a table of splits with multiple metric values per split.

Addi-11:
Hello, I would like to work on this issue. Could our classifier have Non-linear decision boundary??

Addi-11:
Hello, I would like to work on this issue.

Addi-11:
![splits](https://user-images.githubusercontent.com/47007177/76151651-33b8a800-60dd-11ea-82d1-7bfd63f5c74d.png)
Is this the requirement ??

Addi-11:
I am working on the tabulated form and including graphs too. Is anything else required??

Addi-11:
I have submitted a PR regaring this issue, kindly review.

shashigharti:
I will work on this issue

dzeber:
@Addi-11 I saw your PR, we can discuss further there. Yes, the requirement is a function that returns the tabular form.

asthad16:
i will work on this issue

alberginia:
Hi! Yesterday after my pull request I realised that [my solution for issue #53](https://github.com/mozilla/PRESC/pull/72) is actually also addressing this one. I have no experience with git, so I have no clue on how to relate the two issues or how should I proceed so that the pull request is also connected to this issue here.

asthad16:
i have worked on the issue #3. i request u to please review my PR #122

PR #47, opened 2020-03-11
WIP: LR and MLP models for vehicles.csv, please review evaluation methods, unsure about required plots

namrathagopalabhatla:
I am unsure about what plots I should include for the evaluation metrics. Please advise.


namrathagopalabhatla:
I am unsure about what plots I should include for the evaluation metrics. Please advise.

dev/namrathagopalabhatla/classification_models.py
```diff
@@ -0,0 +1,148 @@
...
+from sklearn.metrics import accuracy_score, classification_report, confusion_matrix, precision_recall_fscore_support, f1_score
+from sklearn.linear_model import LogisticRegression
+
+from seaborn import heatmap
+from IPython.display import display
+
+class MLP:
+    def __init__(self, hidden_size=None, activation='relu', solver='lbfgs', max_iter=1000):
+        '''Constructor for initalization of model using provided parameters'''
+        super(MLP, self).__init__()
```
dzeber: This (and line 86) is not needed if your class is not inheriting explicitly from another one.

namrathagopalabhatla:
@dzeber Thank you so much for your suggestions. I will look into it. I am sorry i have been inactive for the past few days. There are a lot of issues that came up with network issues due to the whole pandemic situation.

namrathagopalabhatla:
I have updated the code as per your suggestions. I have also added new code for Issue #3 in my latest commit, where I have added code for AUROC values and ROC curves, and a plot of TP and FP marked separately for each class of the classificarion problem, to visualize the misclassifications in 2D using LDA. Kindly review.

namrathagopalabhatla:
I have updated the repo with the fixes suggested by @dzeber in my first pull request that worked on the startup task #2. I will also add the Eigen value spectrum visualization (Scree plot) in the subsequent commits that also include work on other issues. Kindly review.

namrathagopalabhatla:
I have completed updates according to all the suggestions. I have also incorporated the train-test splits for #3 . Kindly review and tell me if the code for #3 needs anything else

dzeber:
Thanks, your notebook for issue 2 is much improved! It is interesting to see those plots from the PCA - helps explain why those two classes (saab and opel I think) are so difficult to distinguish. Also, good work on the issue 3 notebook - it is very well executed.

PR #51, opened 2020-03-12
issue#2-trained-model-using-svm-for-vehicles-recognition-using-silhouette

shashigharti:
I have trained a model using SVM to detect vehicles using silhouette. There are 4 vehicles categories: 'van', 'saab', 'bus', 'opel'.

**Steps:**

1. Loading the Data
2. Exploratory Analysis
3. Data Preprocessing (Outlier Removal)
4. Hyperparameter Tuning(GridSearchCV)
5. Data Transformation(StandardScaler, MinMaxScaler)
6. 5 Different Experiments : Training and Testing a classifier (SVM - Support Vector Machine)
     - Without Outlier Removal
     - With Outlier Removal  and Various Data Preprocessing and Parameter Tuning
7. Evaluation

PR #57, opened 2020-03-13
issue#3:traversal-of-the-space-of-train-test-splits

shashigharti:
- 20 different passes with different ratio of test-train data.
- Performance grid table and graphs


shashigharti:
Thanks @mlopatka for the review :). I will update the code as per the Python Black formatting guidelines.

And regarding the outlier removal I will definitely use outlier removal methods in my other contributions. I had actually used it in my PR for issue #2 https://github.com/mozilla/PRESC/pull/51 but later I changed it to conservative approach.

shashigharti:
@mlopatka I have applied the formatting to .py file :) . It is ready to be reviewed now. 

Thank you!!

shashigharti:
@mlopatka  As per your suggestion, I have added a new notebook "issue#3-traversal-of-the-space-of-train-test-splits-II.ipynb" where I remove the outliers before running this test.

Thanks!

dzeber:
This is a very nice notebook. Interesting that the results are so different between the two, although the dataset is on the small side (<1000) - it would be interesting to see how much those curves vary across repeated runs. Ie, how much of that is just natural variability. The magnitude of the differences is not very large.

Following up on Martin's comment, I would have gone the other direction and just not removed any points - outliers should generally only be dealt with if they appear to be points that don't belong in the dataset (eg. errors) or would significantly skew the outcomes. The boxplot rule of thumb for outliers was designed more for well-behaved, smaller, more symmetric distributions and does not always work well with modern data.

shashigharti:
Thanks @dzeber for the feedback. I will test that as well. And will follow your suggestion on outlier removal.

PR #60, opened 2020-03-13
WIP: Added KNN classifier for winequality.csv

janvi04:
I have trained and tested KNN model on winequality.csv, I will add some more classifiers on the same dataset. This PR is for initial feedback. Also I will be adding plots to study the relation between the various parameters and the wine quality, so that we can use only those parameters to train and test our model to get a better accuracy.

### DIFF
--- a/dev/janvi04/Classification using KNN on wine quality dataset..ipynb
+++ b/dev/janvi04/Classification using KNN on wine quality dataset..ipynb
@@ -2,7 +2,7 @@
  "cells": [
   {
    "cell_type": "code",
-   "execution_count": 1,
+   "execution_count": 2,
    "metadata": {},
    "outputs": [],
    "source": [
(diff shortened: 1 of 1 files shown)


janvi04:
@dzeber , I have added the code and explanation for KNN algorithm. I am now working with plots to study the relation between quality of wine and other parameters like alcohol fixed acidity etc.
Meanwhile can you please review the code and give initial feedback on this PR.
Thanks.

janvi04:
@mlopatka . Thanks for the review. You are right. The quality column if added in the train set should give us 100% accuracy. Thanks for pointing this out. I will update this soon.

janvi04:
Hi @mlopatka , @dzeber ! I have made the changes. After dropping the quality column, the accuracy of the model decreased to 77% which is less than the previous accuracy. I think now the model is quite accurate as I removed the quality column.
Can you please see this.

Issue #63, opened 2020-03-13
Learning from misclassifications


dzeber:
When training a classification model, it is common to look at accuracy and the confusion matrix, which give a summary view of misclassifications. By itself, these metrics are informative but not very actionable.

Develop a metric or visualization that reveals something more about each misclassified point (beyond just the fact that it was misclassified) that can be used to improve the model.

Some examples of a metric might be the classification probability scores for the different classes, which can indicate whether the misclassified points were close to the decision boundary or not, or the distance from the class mean in feature space, indicating whether the misclassified points are outliers.

A good place to start is to study the misclassifications you got from your model for task #2. What do they tell you about how to improve your model?

PR #64, opened 2020-03-14
 For #63: Learning from misclassification

KaairaGupta:
I completed my first attempt at the problem. My current work can be seen through the examples `example_plot_misclassification_data_binary`, `example_plot_misclassification_data_multiclass`.

The function `plot_misclassified_probablities` in file `plot_misclassification_data.py` has the following specifications:
- Input:
            - model_res: Model result (n X k) matrix where n = number of test_examples, k = number of classes. Each row has probablity values predicted by the model for each of k classes.
            - class_lables: Labels for classes (1,2,3,...k) in order according to model_res.
            - correct_y: Array of n values with each value as label for classes.
- Output: Plot showing classification probablities for misclassified data points.


KaairaGupta:
@dzeber @mlopatka please take a look.
Thanks!

PR #65, opened 2020-03-14
WIP : Learning from misclassifications

Soniyanayak51:
PR for issue #63 
Added plots to visualize distance from decision boundary(from sklearn decision_function) and predicted probabilites for misclassified and correctly classified data for binary classification.

For multiclass classification,  sklearn's OneVsRestClassifier decision_function can be used for SVC.
@mlopatka kindly review


dzeber:
These visualizations are a good start at digging into the way the classifier performs on individual samples. One thing we learn from the plots that wasn't obvious before is that the SVM seems to separate the classes better, as there are two distinct bands compared to the Logisitc.

If you would like to submit this as is, please remove the "WIP" tag from the title and we will merge it. Otherwise, feel free to continue working on this. Either way:

- please format your modules with Black as mlopatka requested
- as in your other PRs, please make sure to use the correct data file in the repo so that others can rerun your code.

For possible next steps, how could you use these plots to improve the performance of your classifiers? A first step that would be helpful is to separate the classes visually, as we currently don't know which class a misclassification belongs to. It might be good to separate correctly classified and misclassified into separate plots. It would also be helpful to write up your interpretations of the plots in the notebook.

Soniyanayak51:
@dzeber thank you for the review. Working on the changes.

dzeber:
Thanks for doing the cleanup. Merging in as is, even though you still have the WIP tag in the title, as it satisfies the issue. If you are planning to continue working on this feel free to open a new PR.

PR #74, opened 2020-03-17
[ Fixes: #63 ]Learn from Misclassification

### LINKED ISSUE #63: Learning from misclassifications

archisha-chandel:
[ Fixes #63 ]

The approach used here gives the user to make the model either more sensitive or more specific depending on the dataset being used. It has been tested on `eeg.csv` where the model is more sensitive.

**Note:** No pre-processing has been performed on the data and any other model with better performance can be used. The aim of this notebook is solely to 'learn from misclassification' and it does not focus on training or improving model performance.


archisha-chandel:
In the notebook there are option for using `precision_score` and `recall_score` as the criteria and the results of both are shown.
1. When taking `precision_score` as the criteria for grid-search:
image
2. When taking `recall_score` as the criteria for grid-search:
image

From this we can see that there is not much improvement in the recall score that is why we move on to the second strategy. Depending on the use case, that is if we want to focus more on precision or recall one of the above two criteria can be used for grid-search. For EEG problem we focus on increasing the recall as this problem requires our model to be more sensitive rather than being specific.

PR #81, opened 2020-03-18
Calibration plots for Defaults dataset

dzekem:
Calibration plots for both Logistic Regression with L1 Regularization and Decision Tree with hyper parameter tuning implemented. 
Calibration being a post-processing technique used to improve a model performance by comparing the actual output and the expected values, meaning the distribution and behavior of the probability predicted should be like that of the actual training data.
This visualization has been appended to the bottom of the jupyter notebook.


dzekem:

dzeber:
This is a nice visualization of the built-in calibration curve values.

PR #84, opened 2020-03-18
WIP: issue #2 Training classification model for vehicles dataset

tasbihaasim:
Hello,

I am getting this error in my Cross Validation function. Can anyone please look at my code and tell me where I am going wrong?

Thank you!

Type_error


tab1tha:
Hello @tasbihaasim I noticed that you created a customised function to perform cross validation. This is not necessary scikit learn has that function already. It is called cross_val_score and you can import it as such ` from sklearn.model_selection import cross_val_score` it splits the dataset into parts and calculates the prediction accuracy per part. It returns a list of all the accuracy values. You can read more about it here https://scikit-learn.org/stable/modules/cross_validation.html .
You must not use cross_val_score, there are other options in the linked article. 

However, if you want to use your own function, 
1. According to the error log, line 28 of cross_validation.py is the source of the problem
2. use iloc in the assignments. For example X_train = X.iloc[train]

tab1tha:
@tasbihaasim also, my text editor showed a notification that accuracy_score is not defined. This is because even though you imported `metrics` from `sklearn`, you have not specified it when calling accuracy score. Instead of `accuracy_score(Y, y_pred)` use `metrics.accuracy_score(y_test, y_pred)`
However, if accuracy_score is the only metric that you will use, then in the import statement, import accuracy_score from sklearn.metrics. In this way, you can simply say 'accuracy_score' when calling it and it will be recognisable throughout the program. 

In summary, these are the changes needed for your code to run properly
1. use `iloc` in assignments as stated before
2. use `metrics.accuracy_score` instead of `accuracy_score`
3. change y_pred[test] to simply y_pred (it's just a variable name)
4.  change the parameters of accuracy_score() from (Y, y_pred) to (y_test, y_pred)

It should work fine, so you can continue.

mlopatka:
More than 5 days inactive; closing out this PR.
If this is still work in progress, feel free to reopen and flag me.

PR #87, opened 2020-03-19
fixed issue #2

ishagarg06:
I worked on winequality dataset. Build the data using various classification model. 
Models used- Decision tree, Random forest regression and Support vector machine
Evaluation is done using confusion matrix and accuracy score.
@dzeber @mlopatka  Please review it once :)
Thanks!

### DIFF
--- a/dev/ishagarg06/WINEQUALITY.ipynb
+++ b/dev/ishagarg06/WINEQUALITY.ipynb
@@ -1,0 +1,1033 @@
+{
+ "cells": [
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "## WINEQUALITY"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "### Import essential libraries"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": 1,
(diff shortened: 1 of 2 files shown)


ishagarg06:
@mlopatka @dzeber Modified the model as per the changes requested for issue#2
--> Path is set for the dataset
--> Quality attribute is discarded.
--> Explained the train-test-split
Please verify it. Thanks!

dzeber:
Thanks for making these updates. One last thing: you don't seem to have used your `winequality_doc.py` module - was this intentional? Ideally you would call those functions and display their outputs in the notebook.

ishagarg06:
@dzeber @mlopatka  Earlier I thought of using `winequality_doc.py` module. But later I called functions and displayed outputs in the notebook itself. If you say I will delete that module.

ishagarg06:
> Thanks for making these updates. One last thing: you don't seem to have used your `winequality_doc.py` module - was this intentional? Ideally you would call those functions and display their outputs in the notebook.

@dzeber @mlopatka Shall I remove that module `winequality_doc.py` or not?

dzeber:
We had requested in the original contribution guidelines that you call the functions from the external module rather than running everything in the notebook. Let's leave it as is at this stage but keep the module for future reference.

PR #88, opened 2020-03-19
Issue#3 traversal of the space of train/test splits

shiza16:
First attempt on Issue#3.
- an evaluation metric is computed with different split ratio of train/test set using repeated K-Fold cross-validation over K values in a range 2 to 5
- and a Line Plot is drawn for each split  with the performance score.
I have two .ipynb file in this, v2.ipynb file is the one that is solely used for issue#3 and i have removed the previous work from this file and another is in which i have merged all the issues that i have worked on so far, Kindly take a look and let me know If i should discard one file or keep the both of them or is there anything i need to change.


shiza16:
Hi @mlopatka , Sorry for the delay but i have made the requested changes. Please review.

shiza16:
I have also moved the file related issue#3 to specific folder to remove the conflicts.

PR #90, opened 2020-03-20
Issue#5-Calibration Plots

shashigharti:
Solution for issue #5 

In this note book, I am comparing 4 different classifiers and to see how calibrated their predictions are i.e the confidence in prediction of each model. 

***Dataset***

I used eeg.csv dataset because this is the binary classification problem and data does not have null values. The dataset has 14976 rows. I am using 70/30 split ratio.  And I have used StandardScaler to transform the dataset.

***Classifiers Used***

1.   Logistic Regression
2.   Random Forest Classifier
3.   SVM (Support Vector Machine)
4.   Naive Bayes


shashigharti:
@dzeber @mlopatka This PR is ready for the review :). Looking forward for your feedback .

Thank you!!

PR #92, opened 2020-03-20
issue #2 Training classification model for winequality.csv dataset

msmelo:


msmelo:
Hello @mlopatka  
Thanks for the feedback! I've made some changes and I've deleted the csv file from the folder. I struggled a little bit to commit the changes, but I believe the file is updated now.

PR #97, opened 2020-03-21
Visualization for misclassifications on default dataset.

dzekem:
MIsclassification must occur during model testing else this will result to model overfitting which occurs when the model tries to be perfect. 
Visualizing this therefore gives a better understanding of the model performance. Misclassification is therefore the total number of wrongly placed data points when the model is performing predictions i.e it is the sum of all False Positive and False negative values.
These visualizations have been considered for both models after training each on the Default data set.


dzekem:

dzekem:
Thank you for your review @dzeber. I have done the changes as you mentioned above.

PR #99, opened 2020-03-21
Comparing different train test ratios

dzekem:
When splitting data to be used for training and testing the model, the ratio of these two sets also influences the model performance especially for small dataset sizes, the ratio differences should not be too different. 
Trying out a ratio of 80:20 and 60:40 for train and test set respectively. The model performs better when the split ratio is 80:20 since it is a large dataset and again the target classes are just two then the model performs better with this ratio.


dzekem:

dzeber:
It doesn't look like you have addressed issue #3 in this notebook. Note that the question
> 60/40 split rather than 80/20
is just given as an example, but the actual task is to write a function that traverses a full grid of train-test splits. I'm closing this PR, but feel free to open a new one if you update your code to address the issue.

PR #101, opened 2020-03-22
Fixed #3, Traversal of the space of train/test splits.

### LINKED ISSUE #3: [Outreachy applications] Traversal of the space of train/test splits

Bolaji61:
Here, I have been able to create a `traversal()` function that takes estimator and dataset as parameters and returns a result of different test sizes and their corresponding accuracy score based on the Logistic Regression estimator used.

However, I do not yet understand what is meant by "a table of splits with multiple metric values per split" as mentioned in issue #3 , I'll appreciate if this is further clarified.


Bolaji61:
I now understand that multiple metrics includes F1-score, Recall, Precision, Classification report, therefore I have added these metrics to each train/test split. However I still have a challenge with finding out a way to improve the view of these metrics.

Bolaji61:
I have fixed issue #3, by investigating and explaining on how much the performance metrics/scores depend on the choice of train/test split proportions using Logistic Regression model

PR #105, opened 2020-03-22
Learning from misclassifications 

dzekem:
A simple and straightforward approach to improving the performance of a classifier that predicts probabilities on an imbalanced classification problem is to tune the threshold used to map probabilities to class labels. This can be done by using ROC Curves and Precision-Recall Curves, the best or optimal threshold for the classifier can therefore be calculated directly.
  This approach has been used to learn from the misclassifications using a Logistic regression model and then optimized.


dzekem:

PR #106, opened 2020-03-22
Importance score for dataset training on defaults dataset

dzekem:
This notebook brings out the difference of training a model with a record(data point) removed from the train set and training the model again with the record maintained.
After dropping this record, this caused the accuracy score for the test set to increase from 0.813 to 0.8145. 
this brings an accuracy difference of 0.0015 which shows this record has a 0.15% influence on making the model better.


dzekem:

dzekem:
@mlopatka Thank you for the review. I have implemented it with evidence that the record is from the train set

dzekem:
Thank you for the review @mlopatka I have considered expanded on it by considering 6 different data points and accuracy score printed out when each is dropped with the difference. All this visualization has been done on a table.

mlopatka:
@dzekem thanks for the quick reply. I don't understand why these 6 specific points are a generalizable sample for the entire data set. 

Is there something specific about: `drop_records = [64, 159, 181, 371, 377, 452, 481]`?

Is there any reason not to evaluate the performance when *every* point is taken out on one iteration? You could then accumulate the differences in performance so that you can plot the distribution of differences in performance. 

This is strongly related to an analysis of the performance across a varying number of cross-validation folds (issue #4) and a comparison of those two distributions would be interesting. 

As it is, I am not sure the motivation for selecting these specific points, or the generalizability of these findings.

dzekem:
@mlopatka please I don't seem to understand the comment above. This PR is for issue #8 which was stated that it could be resolved by  "... Implement a way to assess the importance of an inidividual training datapoint to the performance of the model. This could be done by training the same model with a particular point included and then excluded from the training set, and computing the difference in performance scores on the test set ..." 

This is what I have implemented in the notebook above. To show that a particular data point belongs to the train set after the data has been split and also the influence of this data point to the test accuracy when dropped.

The only way I could take note of a data point that falls under the train set after splitting the data was to convert the numpy array Xtrain to a dataframe and then do an inner join with the initial data to see records that they have in common. 
This join is done with respect to the feature "ID" from which the selected records were displayed as common records that fall under the train set. 

`common = pd.merge(trial.head(500), trial_X.head(500), how='inner', on=['id'])`

The  `drop_records = [64, 159, 181, 371, 377, 452, 481]` are therefore indexes that corresponds to the 6 records with IDs displayed on the table after running the inner join above.

To see the influence of each data point on the test accuracy when dropped, I used a for loop to loop through all of them and also append the results in the variable score.

mlopatka:
I understand that this is a solution for issue#8.

With the comment above, I meant to communicate that issue #8 and issue #4 are related in that they both ask for different ways of investigating the *stability* and *generalizability* of performance observed during the training of a classifier. 

I understand your approach, but I do not know why the results based on the 6 specific points for which you show results should be expected to hold for the rest of the data set. One possible approach would be to drop a single point from Xtrain _after_ the split. This would allow you to re-train and re-test the classifier many times, each  time dropping a different point from Xtrain. If you do this for every point in Xtrain, we start to get an idea of the variation that any single point can be responsible for. 

Then the question becomes, whether there was anything special about that initial split into train/test. And that is the separate (but related) topic that issue #4 is meant to address.

Does this help clarify my comments above @dzekem ?

dzekem:
Not yet clear Sir.

The data points dropped before the split will still be the same after the
split since I used a random state of 0 all through out the notebook.

The initial split above using X_try was for me to bring out those data
points that at any moment when the data is split, those records will fall
under the train set.

And again after splitting a dataframe, the different splits have data types
numpy arrays which has to be converted back to dataframes in order to be
able to yo better visualize the data and also manipulate it with ease.

The first split above was a way for me to bring out those data points and I
stated a random state of 0 which has been the same through out the
notebook.

I don’t think the approach you are proposing below is different from my own
implementation as wether the record is dropped before or after the split,
data points already verified that falls under the train set will still work
since during the testing, random state was set to 0 which was still
maintained everywhere in the code thereby all the random variables placed
under the train or test sets will always be the same after any number of
computation time.

mlopatka:
Ok, let's focus on the main issue.

Why use only 6 points?

Instead, can you use your current approach, by expanding `drop_records` to equal `Xtrain`, or even just using the `for` loop to iterate over `Xtrain` like this: `for i in Xtrain:`. 
Then the `New test accuracy` and `Difference` columns of your output DataFrame will contain a large number of samples that capture the variation we see in performance depending on which data point is removed.

That distribution (from the whole column of `Difference` values) is interesting when solving issue #8. A nice way to wrap up this notebook would be to plot it like a histogram, for example.

dzekem:
Thank you for the clarification @mlopatka 
I have changed implemented it using the approach you mentioned above. Since the dataset size was too large(24,000 records), I tried using all and my machine took hours and still wasn't able to compute the output which I then had to force stop and then reduce the size.
 I instead used the first 200 in the train set dropping each and calculating the test accuracy and accuracy difference.
After which the difference was plotted on a histogram which shows non of the records after being dropped caused an increase in the test accuracy score instead they all caused a fall in the accuracy score sum with high values like 0.009.

Hope I got it right this time :)

mlopatka:
Awesome! I ran your code locally for all 24000 rows (it did indeed take about 3.5 hours) but it looks like the results are consistent with what you saw for the first 200.

Thanks for addressing the feedback. I would encourage you think about why the extension to a large sample is important as we explore the stability of performance.

I'll add a commit with the rendered notebook from my local work and merge this in.
Great work!

PR #107, opened 2020-03-22
comparing test samples classifications between two models

dzekem:
Making predictions using two models to see which has better performance is a vital step in machine learning. Comparing the prediction probabilities of both models used here gives a better understanding of the model performance as a visualization brings out the different probabilities considering the class 0(Non defaulted credit cards) and class 1(defaulted credit cards)


dzekem:

dzekem:
Thank you for your review @dzeber. I have implemented the changes you mentioned above

PR #110, opened 2020-03-22
Calibration plot [fixes # 5]

### LINKED ISSUE #5: [Outreachy applications] Calibration plot
A calibration plot assesses how well the predicted class probabilities match the actual rate of occurrence in the dataset.

Write a function to create a calibration plot given one or more binary classifiers and a test set. For each model it would compute predicted probabilities for each test datapoint. These probability values should then be binned according to a scheme (eg. intervals of 10%), and the observed occurrence rate can be computed as the proportion of true positive class samples out of all samples in each bin. The plot then displays the observed occurrence rates vs the bin midpoints for each model.

Sidrah-Madiha:
this branch fixes #5 , there is a helper file calibration_plot.py, it plots calibration plot for multiple classifiers and also plots perfectly calibrated line as a reference to see how close are predicted probabilities are to the true probabilities. A demo of the graph is also shown with its interpretation in Test_calibration_plot.ipynb. 
Calib


Sidrah-Madiha:
@mlopatka  fixed conflicts, please review

PR #113, opened 2020-03-23
#7 Visualization for misclassification

BBimie:
Hi, I was able to do a visualization of the distribution of mis-classified data points in the test set, but I need guidance with plotting the prediction probability against the different data points.


dev/BBimie/misclassification_function.py
```diff
@@ -0,0 +1,82 @@
...
+        
+        #get the prediction probability and make it a dataframe
+        probability = clf.predict_proba(a)
+        probability_df = pd.DataFrame(probability)
+        
+        #get the maximum probability for each prediction
+        max_prob = probability_df.max(axis=1)
+        
+        #change the indexing of the test set to start from zero
+        test.index = np.arange(0, len(test))
```
dzeber: You can also use `test.reset_index(drop=True)`.

dev/BBimie/misclassification_function.py
```diff
@@ -0,0 +1,82 @@
...
+        #train target
+        y = train[target]
+    
+        #test feature
+        a = test.drop([target], axis=1)
+        #test target
+        b = test[target]
+    
+        #fit the random forest model
+        clf.fit(x,y)
```
dzeber: Rather than doing train/test splits and fitting the model inside this function, it would be nicer to have a dedicated function for computing the probabilities and creating the visualization that take as input a fitted model and dataset.

dev/BBimie/#7 visualization for misclassifications.ipynb
```diff
@@ -0,0 +1,1032 @@
...
+  },
+  {
+   "cell_type": "code",
+   "execution_count": 4,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "current_dir = os.path.abspath(os.getcwd())\n",
+    "\n",
+    "dataset_path = os.path.join(current_dir, '..', 'datasets', 'winequality.csv')"
```
dzeber: Please fix the dataset path here and in the other notebook (needs an extra `..`)

PR #115, opened 2020-03-23
covariate shift with Bike sharing demand dataset

dzekem:
Covariate shift occurs when two data sets are joined(inner join) when they had some time difference thereby making the content of the train and test datasets different. 
Training a model gives a different accuracy on the train and a larger difference is observed on the test accuracy. This is due to data drifting which when noticed could either be dropped or implementation od importance weight using density ratio estimation


dzekem:
@mlopatka and @dzeber please can you review my PR. I'm already done with all the issues and reviews on these PRs will help me make final application stage for this project.

dzekem:
@mlopatka please this notebook has not yet been reviewed.

PR #119, opened 2020-03-24
issue#8-importance-score-for-dataset-training-samples

shashigharti:
* Importance score function to calculate sensitivity of model to the data points.
* Visualize the negative and positive impact of the data points(in cluster) to the overall accuracy.

@dzeber @mlopatka this is ready to review :)

Thank you!!


dzeber:
Nice work! I like the graphical presentation, and it's interesting that the accuracy changed so much after removing those points.

One thing to note (but not requested as a change at this time) is that you are re-splitting train and test for each point, meaning that even if accuracy is not actually different after removing the point, there will be some natural variation in the accuracy scores reported. With this approach we might want to provide a baseline expected variation measure, or consider other splitting strategies (eg. something like deterministic folds).

shashigharti:
Thank you @dzeber

PR #26, opened 2020-03-08
wine quality classification using decision tree binary and multi-class classification

asthad16:

PR #122, opened 2020-03-25
Asthad16 issue3 train test split

asthad16:
this PR fixes issue #3 . it covers following things:
1-this uses KNN model as a part of further recommended step of using outlier detection algorithm for classification in previous PR #26 .

2-it performs accuracy evaluation of model with different train-test split ratio.

3-also further hyper parameter tuning is done and the overall accuracy obtained from KNN is improved than obtain from decision tree in #26 used in issue #2


asthad16:
the notebook and modules in this PR uses modules of verious PR #26.

PR #123, opened 2020-03-25
Asthad16 #4 kfold

asthad16:
this notebook and module uses the previous modules of PR  and fixes issue #4 .


dzeber:
Thanks for this contribution exploring the space of cross-validation folds. Merging this in as it fulfills the task, although there is still room for improvement. Ideally the model would be passed as a function arg, rather than hard-coding it into the function, so that it can be applied to any general model. Also, it would be really interesting to include some larger fold sizes for comparison. Eg. a well-known approach is Leave-one-out CV, which is n-fold CV.

asthad16:
thanks a lot for reviewing it i will make a new pull request and work on
your suggestions.

PR #124, opened 2020-03-25
WIP: #4 Cross fold validation task complete 

namrathagopalabhatla:
I have completed the code for #4 and think it will solve the issue sufficiently. I also have other commits for multiple other issues, but by mistake have put all that code in my previous pull request #47 . Please review the changes made and the issue resolved in those as well. I apologise for the error on my part, and I'll make sure the PRs for different issues stay on different PRs here on. 

Also, this commit converts all code to be python black adherent.


dzeber:
Thanks, this is a nice solution for this task. The code is clean and documented and I like that you tried different grid configurations for traversing the possible fold sizes and left that as a function param.

This could be improved by making it a bit more general (not requested at this time): passing the classifier in as a function param rather than hardcoding it in the function, and trying a few larger fold sizes, eg. n-fold (leave-one-out CV).

PR #125, opened 2020-03-25
WIP: Visualization of evaluation metrics

namrathagopalabhatla:
The visualization code and notebook has been added, please review and suggest any improvements necessary. Different distributions of randomly generated toy data have been used to show the visualization function.


dev/namrathagopalabhatla/visualize_eval_metric.py
```diff
@@ -0,0 +1,74 @@
...
+def generate_data(n=15, dist="gaussian", scale=0.5, prob=0.2):
+    """Function to generate data of different types of distributions"""
+    if n < 3:
+        print("Too few points, minimum n required is n > 2.")
+        return None
+
+    X = np.arange(15)
+    data = []
+
+    for i in range(len(X)):
```
dzeber: This could just be `range(15)`, and then `i` instead of `X[i]` on line 21. Although, really the `15` (number of rows) should be passed as a function arg.

PR #129, opened 2020-03-25
WIP Issue #6 visualization of evaluation metric

urvigodha:
Hi @mlopatka , @dzeber please take a look at my attempt at issue #6  thanks!


dzeber:
This is a good start. Please move the code to generate the plot to a function in a separate module. We can merge at that point to satisfy the task in the issue. That said, it would be nice to try this out on an actual evaluation metric for a classifier.

PR #131, opened 2020-03-25
Traversal of train test splits

SaumyaSinghal:


dev/Saumya/Train-test split traversal.ipynb
```diff
@@ -0,0 +1,1638 @@
...
+    "    result_list = []\n",
+    "    for i in range (start, end, interval):\n",
+    "        result_dict = {}\n",
+    "        test_ratio = i/100\n",
+    "        \n",
+    "        result_dict[\"Test Ratio\"] = test_ratio\n",
+    "        \n",
+    "        train_X, test_X, train_y, test_y = model_selection.train_test_split(X, y, test_size=test_ratio, stratify=y, random_state=42)\n",
+    "        train_X = (train_X - train_X.mean())/train_X.std()\n",
+    "        test_X = (test_X - test_X.mean())/test_X.std()\n",
```
dzeber, line 194: You should be normalizing the test set using the mean and std of the __training__ set, same as for the training set itself. Thinking of the normalization parameters as part of the model, this ensures that your model is not learning anything from the test set. In fact you can use the built-in [StandardScaler](https://scikit-learn.org/stable/modules/preprocessing.html#standardization-or-mean-removal-and-variance-scaling) to handle this.

dzeber:
Actually, let me merge this in right away, so that you can record this contribution, and you can submit an additional PR to fix the scaling issue if you have time.

PR #133, opened 2020-03-25
WIP: Visualize misclassifications

SaumyaSinghal:


dzeber:
What I was really hoping for with this task was to go beyond the built-in plots like confusion matrix and ROC curves to learn something about the characteristics of the specific points that were misclassified. For example, were they a near miss from the classification boundary, or was the model way off?

However, your notebook is well-presented and your analysis is thorough, so I will merge this in as satisfying the issue. Also, note that `plot_confusion_matrix` and `roc_auc_score` are not included or have different signatures in the version of scikit-learn in our environment - you should really be running your code in the conda environment provided in the repo.

Issue #78, opened 2020-03-18
[Outreachy applications] Covariate Shift


archisha-chandel:
As said in issue #3, a function has been written to check the evaluation metric for a model when train-test split ratio is varied from 0.0- 1.0. But there are cases where we cannot just randomly decide the split ratio on the basis on evaluation metric as the metric does not improve beyond a certain point. Example: [data set](https://www.kaggle.com/c/porto-seguro-safe-driver-prediction/data) based on predicting the probability that an auto insurance policy holder files a claim by Porto Seguro has data which has covariate shift. For this data using the regular methods does not yield a good evaluation metric score.

Propose a solution to detect and handle covariates in the data. The above mentioned data or any other can be used for implementation which may or may not have covariates.
**Note: This issue has been created to provide a solution to issue #3 which has an exception of not dealing with data that has covariate shift.**

Covariate Shift: It refers to a situation where predictor variables have different characteristics (distribution) in train and test data. So our model is unable to predict well on the test data as it has been trained on a certain set of feature characteristics which are different from test data. 

![image](https://user-images.githubusercontent.com/47753303/76925232-c12aa200-68fe-11ea-9e8c-b2e2572f0192.png)

namrathagopalabhatla:
Are we supposed to work on this issue?

mlopatka:
You are free to work on this issue, but it is not required or expected.
You can also spend the next days finalizing your outreachy application.

PR #80, opened 2020-03-18
[ Fixes: #78 ] Covariate Data

archisha-chandel:
[ Fixes: #78, #3 ]

- StratifiedKFold and RandomForestClassifier has been used to detect covariates in data.
- Improving performance by assigning importance weight using Density Ratio Estimation
- Evaluation metric score before covariate analysis: 0.52
- Evaluation metric score after covariate analysis: 0.96
- False Negatives can be reduced using #74 
image

PR #136, opened 2020-03-29
[ Fixes: #78 ] Covariate Data

archisha-chandel:
[ Fixes: #78, #3 ]

- StratifiedKFold and RandomForestClassifier has been used to detect covariates in data.
- Improving performance by assigning importance weight using Density Ratio Estimation
- Evaluation metric score before covariate analysis: 0.52
- Evaluation metric score after covariate analysis: 0.96
- False Negatives can be reduced using #74 
image

- The notebook itself downloads the data and stores it in datasets folder (Run cell 1)


archisha-chandel:
**This PR is made as an extension to #80 which could not be fixed due to some mistake made and as a result had to be closed**

PR #155, opened 2020-04-05
Visualization for misclassifications

mhmohona:
This notebook is created to do tasks given in the issue **Visualization for misclassifications**.


mhmohona:
I was confused how to split tasks for `Visualization for misclassification` and `learning from misclassification` and ended with doing almost same tasks for two issues. Now I got another way to approach  `learning from misclassification` task, so updating `Visualization for misclassification`  by merging my work from these two notebooks.

PR #158, opened 2020-04-05
general  train-test split and k-fold evaluation

asthad16:
though the initial PR made were successfully merged but they were not general for any model. now i have generalized them and also added
another variant of k-fold cross validation (LOOCV) for improved results.


asthad16:
@dzeber u suggested me to implement leave one out CV and also my function was not generalized for any model for #3 and #4 issues but now i have fixed this.

asthad16:
please do consider #158 and review it.

Issue #170, opened 2020-04-20
Multi-Class Imbalanced Data


archisha-chandel:
Recent studies point to the fact that it is not the imbalanced data itself, but rather other data difficulty factors, amplified by the data imbalance, that pose a challenge during the learning process. Such factors include small sample size, presence of disjoint and overlapping data distributions, and presence of outliers and noisy observations. Multi-Class nature of classification can further amplify the problem if the data is imbalanced.

Earlier suggestions for dealing with multi-class imbalanced data involves decomposing the given problem to binary. This leads to quite a significant loss of information and loss of relationship among the classes being decomposed. For instance, it is possible that `class A` can be a part of the majority for `class B` but be a minority for `class C`. Real-world scenarios have more complex problems.

**Propose a solution to deal with imbalanced data problem with embedded data-level difficulties, i.e., atypical data distributions, overlapping classes, and small disjuncts, in the multi-class setting.**

PR #171, opened 2020-04-20
[ Fixes: #170 ] Multi-class Imbalance Data with Noise

archisha-chandel:
[ Fixes: #170 ]

The main idea for proposing a new algorithm is:

- A smart oversampling scheme that makes use of all the information including the interrelationship among different classes.

- Ability to handle imbalanced data with label noise.

- To provide a reliable trade-off between processing time and obtained improvements.

- To overcome the drawbacks of SMOTE-based methods.

**Input:** `python prepare_resampled_data.py -dataset zoo -partition 3 -fold 1`
The `-dataset`, `-partition` and `fold` values can vary. There are other parameters that can be adjusted as well.
image

**Output:** 
image
The output explains the `label : count of instances with that label` before and after performing the algorithm. This would create a CSV file in the resampled_data directory with oversampled vehicle dataset (first partition and first fold; data was partitioned using 5x2 cross-validation, as described in the paper).

**The main steps are:**

1. Create a sphere around minority data points
2. Clean-up majority data points inside the sphere
3. Adaptive oversampling based on sphere radius

**Step 1:** 
image
**Step 2 and 3:**
image
Finally, you can import the algorithm itself from the algorithms.py file and use it in your own code: it uses the API proposed by imblearn package, so you just have to call the `fit_sample` method.

The above approach is based on https://github.com/michalkoziarski/MC-CCR obtained from the paper published on 7th April, 2020 - https://arxiv.org/pdf/2004.03406.pdf.


mlopatka:
@dzeber can you take a peak at this PR please?

PR #177, opened 2020-07-08
Add PyPI upload support

crankycoder:
* twine has been added as a dependency to support PyPI uploads
* `upload` target added to Makefile
* PyPI upload docs added to README
* setup.py modified to include a long_description for the PyPI index

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -1,14 +1,21 @@
 from setuptools import find_packages, setup
 
+
+with open("README.md", "r") as fh:
+    long_description = fh.read()
+
 setup(
     name="PRESC",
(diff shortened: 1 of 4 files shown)


setup.py
```diff
@@ -1,14 +1,20 @@
 from setuptools import find_packages, setup
 
+
+with open("README.md", "r") as fh:
+    long_description = fh.read()
```
birdsarah, line 5:
```suggestion
long_description = open("README.md").read()
```
python got smarter and I like this new shorter way of reading in files.
this could even be inline on line 16. it's not important though.

PR #178, opened 2020-07-08
Add flake8 and black on pre-commit hook

### LINKED ISSUE #103: Make all PRs Black adherent
As mentioned in the [Contribution guidelines](https://github.com/mozilla/PRESC#contribution-guidelines) :
Code formatting guidelines should strinctly adhere to [Python Black](https://pypi.org/project/black/) formatting guidelines.

Now, sometimes the contributors forget to ensure that the above is followed. To reduce the efforts while making sure the required code formatting is followed, adding Python Black as a pre-commit hook would make things easier.

crankycoder:
This resolves #103


.flake8
```diff
@@ -1,4 +1,5 @@
 [flake8]
-max-line-length = 80
+ignore = E203, E266, E501, W503, F403, F401
+max-complexity = 30
+max-line-length = 79
```
birdsarah, line 4: 79 seems very ungenerous - I like 120 these days. IS there any particular motivation for this stringency?

presc/alberginia/missclassification_visuals.py
```diff
@@ -64,14 +64,14 @@ class to use for validation of the model.
 
     elif new_classes == "which-hit":
         dataset_missclass["Miss & Class"] = np.where(
-            dataset_missclass_calc["Hit"] == False,
+            not dataset_missclass_calc["Hit"],
```
birdsarah, line 67: Do you want any "not" value or specifically False?

crankycoder: @alberginia ^

.flake8
crankycoder: Happy to bump this up to 120 in 500e894

presc/alberginia/missclassification_visuals.py
alberginia: @crankycoder @birdsarah Here I wanted to detect the "False" values of "Hit", because thise means that the point was labelled as misclassified. I think that "not" would potentially capture other things, right? Although I don't think there are other things apart from "True" or "False" in this dataset.

birdsarah: Then `dataset_missclass_calc["Hit"] is False` is the pythonic way to be specific to `True/False` as opposed to allowing `1/0` and other things.

alberginia: Thanks! There probably will be a ton of those things.  :roll_eyes:

birdsarah: All good. That's why code review is great. There's always new stuff to learn. Or in my case today: new stuff to relearn after unlearning if (thanks @crankycoder).

alberginia: Actually... I've found out now that this change breaks the function, so I'll have to revert it back to what it was originally.  :P

PR #179, opened 2020-07-08
Reorganising a bit and adding two references

alberginia:


dzeber:
@alberginia Should we also delete the `presc/alberginia` folder, so that it gets moved rather than copied?

alberginia:
I think we need the modules for the misclassification bit. But we can delete the "alberginia" folder and keep the files inside the new "misclassifications" folder.

dzeber:
@alberginia Agreed - so could you delete the `presc/alberginia` from this PR, and then the `presc/misclassifications` will get added when you submit a PR for that work.

alberginia:
@dzeber, I tried to remove the files that @crankycoder moved, as you suggested, but it claims there are complex conflicts that must be resolved. Any idea how to sort it out?

dzeber:
Conflicts are fixed in #180

PR #181, opened 2020-07-23
Misclassification rate as a function of feature values. Functions, tests, and examples notebook. v0

alberginia:
Hello,

I am adding the following new functions for analysis and visualization of the misclassification rate as a function of the features:

* misclass_rate_feature(test_dataset, test_dataset_misclassifications, feature, bins=10)
* show_misclass_rate_feature(test_dataset, test_dataset_misclassifications, feature, bins=10)
* show_misclass_rates_features(test_dataset, test_dataset_misclassifications, bins=10)

The binning in these functions is completely free.

These other three are to compute the appropriate binnings for quartiles, deciles, etc. and also to visualize them and make sure they are good enough.

* compute_tiles(dataset, feature, tiles=4)
* show_tiles_feature(dataset, feature, tiles=4)
* show_tiles_features(test_dataset, tiles=4)

I am also incorporating to the main repo my function from visualization of misclassifications from the contribution period:
* predictions_to_class(X_test, y_test, y_predicted, new_classes="which-fail")

I also added a test file with some tests for the previous functions (except the ones that display figures). I need help making sure that the tests work (locally, they only did in the module folder). 

I also added a notebook with examples of the options.

I also restored the function we "fixed" in #178, turns out we broke it.  :)


tests/test_misclassifications.py
```diff
@@ -0,0 +1,106 @@
...
+
+    edges_bins = compute_tiles(dataset, feature=selected_feature, tiles=tile_number)
+
+    assert edges_bins[0] == min(dataset[selected_feature])
+    assert edges_bins[-1] == max(dataset[selected_feature])
+    assert len(edges_bins) == tile_number + 1
+
+
+def test_show_tiles_feature():
+    pass
```
crankycoder, line 102: Do these need to be filled out before we merge this PR?

alberginia: My idea was that this was supposed to be self-contained, so I built minimal datasets with a similar structure and it shouldn't be necessary for these simple kinds of tests. But I have no experience with testing, perhaps you think more advanced testing should be necessary?

alberginia: Ah, yes, you're right. I didn't because I wasn't sure how to build test for those or even if it was expected. Maybe @mlopatka or @dzeber have an idea of what to do there.

tests/test_misclassifications.py
```diff
@@ -0,0 +1,106 @@
...
+
+    result = misclass_rate_feature(
+        dataset, dataset_misclass, feature="Feature 1", bins=bin_number
+    )
+    assert len(result) == 3
+    assert len(result[0]) == len(result[1]) + 1
+    assert len(result[1]) == bin_number
+
+    assert max(result[1]) <= 1.0
+    assert min(result[1]) >= 0.0
```
dzeber, line 70: Since the dataset is so small, you can just check the actual values of the result. Maybe pick numbers so that the rates are nice decimals. The [numpy testing utils](https://numpy.org/doc/stable/reference/routines.testing.html) may also come in handy here.

tests/test_misclassifications.py
```diff
@@ -0,0 +1,106 @@
...
+    )
+    dataset_misclass = pd.DataFrame(
+        {
+            "Feature 1": [1, 2.5, 5],
+            "Feature 2": [1, 6, 9],
+            "Miss & Class": ["True", "False", "True"],
+        },
+        columns=["Feature 1", "Feature 2", "Miss & Class"],
+    )
+    bin_number = 2
```
dzeber, line 60: It's a good idea to test all of the main code paths for the function. Here you could do three cases: `bins` is an int, `bins` is an array, and `bins` are quantiles.

tests/test_misclassifications.py
```diff
@@ -0,0 +1,106 @@
...
+    )
+
+    selected_feature = "Feature 1"
+    tile_number = 2
+
+    edges_bins = compute_tiles(dataset, feature=selected_feature, tiles=tile_number)
+
+    assert edges_bins[0] == min(dataset[selected_feature])
+    assert edges_bins[-1] == max(dataset[selected_feature])
+    assert len(edges_bins) == tile_number + 1
```
dzeber, line 98: Similarly check the actual value, eg. `edges_bins == [1, 3, 5]`

tests/test_misclassifications.py
dzeber: Cf. the comment for `test_show_misclass_rate_feature`. The best thing is probably just make sure the function being tested runs without error.

tests/test_misclassifications.py
```diff
@@ -0,0 +1,106 @@
+from presc.misclassifications.misclassification_visuals import *
+from presc.misclassifications.misclass_rate import *
```
dzeber, line 2: Please import the individual functions you will use, even if it's a long list. This creates confusion, since in this case it's also importing `pd` for pandas, and the tests are running without error even though `pandas` is not imported here, so it's hiding the fact that this module also requires pandas.

presc/misclassifications/misclass_rate.py
```diff
@@ -0,0 +1,310 @@
...
+            * If an integer, it divides the feature scale in regularly spaced 
+            bins (default value is 10).
+            * If the string "quartiles" or "deciles" is used, then it 
+            automatically computes the appropriate bin edge postions to optimize 
+            for a quartile grouping. 
+            * If any other feature intervals are needed, then a list of the 
+            feature values corresponding to the positions separating the bins 
+            and including the outermost edges must be provided. 
+    """
+    # Computes position of bin edges for quartiles or deciles
```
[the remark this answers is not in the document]
alberginia: Sorry, i don't understand what you want to remove here.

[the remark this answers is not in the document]
dzeber: Checking the `bins` arg for `"quartiles"`/`"deciles"` will happen anyway inside `misclass_rate_feature()`, so I wouldn't run this code again here - just pass `bins` into `misclass_rate_feature()` as is.

tests/test_misclassifications.py
```diff
@@ -0,0 +1,72 @@
...
+    show_misclass_rate_feature,
+    show_misclass_rates_features,
+    compute_quantiles,
+    show_quantiles_feature,
+    show_quantiles_features,
+)
+
+
+@pytest.fixture
+def test_dataset():
```
dzeber, line 15: This is kind of a nit, but convention is that functions that run tests are named `test_....`, so fixtures shouldn't be. You can just call this `dataset`.

tests/test_misclassifications.py
```diff
@@ -0,0 +1,72 @@
...
+    assert len(result) == 3
+    assert len(result[0]) == len(result[1]) + 1
+    assert len(result[1]) == bin_number
+
+    assert max(result[1]) <= 1.0
+    assert min(result[1]) >= 0.0
+
+
+def test_show_misclass_rate_feature(test_dataset, test_dataset_predictions):
+    show_misclass_rate_feature(test_dataset, test_dataset_predictions, "Feature 2")
```
dzeber, line 50: This is good - but when I run pytest this is opening a graphics device window and showing a plot, and it blocks the test from continuing until the window is closed. This needs to be handled in such a way that the plot isn't acutally displayed (not sure how offhand, probably some pyplot setting).

alberginia:
@dzeber @mlopatka @crankycoder Flake8 doen't seem to complain to me anymore, in case you want to merge this.  :)

mlopatka:
w00t!

PR #184, opened 2020-08-18
Added exploration of the effect of the k-folds on the cross-validation score and variation

alberginia:
I have added here the functionality to explore the impact and stability on metrics (average and standard deviation) of carrying out variations of the number of k-folds when performing cross-validation.

There is one new function:

* **explore_cross_validation_kfolds** (dataset, pipeline, metrics=["accuracy"], kfolds_list=[3, 4, 5, 7, 10], repetitions=1, minimum_repetitions=False)

The user can specify any scaler and model through the "pipeline" parameter, and at the moment only the exploration of the "accuracy" is implemented, but the function is prepared to work with a list of metrics.

For better statistics, it is also possible to specify a fixed number of times to repeat the cross validation for all k-fold values, or a minimum number of total iterations so that the cross validation is only repeated for small k-fold numbers.

The results can be visualized with the already existing function "show_averages_and_variations".

I have also added a notebook with examples and a couple of tests for the function.


presc/training_stability/cross_validation.py
```diff
@@ -0,0 +1,121 @@
...
+    dataset,
+    pipeline,
+    metrics=["accuracy"],
+    kfolds_list=[3, 4, 5, 7, 10],
+    repetitions=1,
+    minimum_repetitions=False,
+):
+    """Explore the cross-validation k-fold number with the specified pipeline.
+
+    Choosing the number of k-folds with which to carry out the cross-validation
```
dzeber, line 18: I see you've got these paragraphs in the notebook as well, and I think they belong there rather than here. Perhaps replace with a short description of what this function does.

presc/training_stability/cross_validation.py
```diff
@@ -0,0 +1,121 @@
...
+        # Gather all repetitions of each kfold
+        scores_kfold = {metric: [] for metric in scores}
+
+        for element in scores_list:
+            for metric in element:
+                scores_kfold[metric] += list(element[metric])
+
+        # Consolidate all repetitions
+        scores_kfold = {
+            metric: [np.mean(scores_kfold[metric]), np.std(scores_kfold[metric])]
```
dzeber, line 109: If we were running CV without repeating, `scores` gives you the score values within each fold, and then we would average. The goal here is to assess how much that average can vary across the repetitions. To do that, we would need to compute the average score for each rep and then find the SD of those averages, rather than the SD across all indidivual fold scores. The difference being that the scores across folds within one rep are not independent.

Consequently, running this function with 1 rep would give us the trend across different fold numbers but no estimate of variability. I recommend setting the default rep value to at least 10.

presc/training_stability/cross_validation.py
```diff
@@ -0,0 +1,121 @@
...
+
+    repetitions_original = repetitions
+
+    scores_summary = None
+
+    for num in kfolds_list:
+
+        # If number of k-folds is smaller than the minimum number of
+        # repetitions, repeat the cross-validation for that k-fold number as
+        # many times as necessary.
```
dzeber, line 83: Cf my comment below, since we want to compute variability across reps, I'm not sure we would want to do this.

PR #187, opened 2020-08-28
Add roadmap doc

dzeber:
@mlopatka @alberginia Migrating the roadmap over from previous drafts. This includes all the same topics but thinking of them more as code features.


docs/ROADMAP.md
```diff
@@ -0,0 +1,442 @@
...
+=====================
+
+This is an overview of the features planned for integration into PRESC.
+It is intended to give a high-level description of how these will work and
+sample use cases.
+Prioritization and implementation details are maintained in the repo
+[issues](https://github.com/mozilla/PRESC/issues).
+
+At the core of PRESC is a collection of analyses that can be run on a given
+statistical model and dataset to inform the developer on different aspects of
```
mlopatka, line 11: nit: "Or combination of model and dataset".

Phrasing could be a bit more direct here. This is the one sentence that contextualizes all that comes afterwards in terms of what the goals of the PRESC project are.

docs/ROADMAP.md
```diff
@@ -0,0 +1,442 @@
...
+the model's performance and behaviour.
+The two main intended uses are a graphical presentation in a report and a
+pass/fail determination obtained by comparing the results against threshold
+values determined by the user.
+In either case, it would be up to the user to decide on a course of action to
```
mlopatka, line 16: I like the emphasis on developer/consumer agency, but we should also emphasize that it is the *responsibility* of the developer to interpret the outcome of these evaluations in a manner appropriate with their domain. PRESC should not promise to be a guarantee against bias, but rather a toolkit for helping produce BETTER ML solutions.

docs/ROADMAP.md
```diff
@@ -0,0 +1,442 @@
...
+correct deficiencies in the model surfaced by these evaluations.
+
+Planned analyses are described below, grouped by theme.
+Some of these will lend themselves to multiple possible visualizations or
+summaries, while others will be applicable in a single clear way.
+As of right now, the details are fleshed out to varying extents.
```
mlopatka, line 22: remove this line:
"As of right now, the details are fleshed out to varying extents."

docs/ROADMAP.md
```diff
@@ -0,0 +1,442 @@
...
+- Selection of number of CV folds to minimize bias and variance
+- Deviation from average-case performance
+
+__Type__: Model performance confidence metric
+
+
+Other
+-----
+
+### Novelty
```
mlopatka, line 407: Structurally, I  think the contents of the entire Novelty section can be integrated under other themes. 
the comparison as a function of time is not fundamentally different that a comparison between multiple models trained over different subsets of available data contemporaneously available. 

I really like the clarity of the themes wihtout the need for a catch-all "other" category. Perhaps we can either roll these contents into their appropriate sections or just exclude this section for now?

dzeber:
Thanks for the feedback - I've made some updates to address the comments. I moved the Novelty section back to the Robustness theme, where it was in the original draft. I think the only reason I broke it out was because I felt the approach was quite different and not model-specific, but you're right, the distinction feels artificial.

Issue #188, opened 2020-09-10
Implementation for conditional metrics


dzeber:
Implement the computation for [conditional metrics](https://github.com/mozilla/PRESC/blob/master/docs/ROADMAP.md#conditional-metrics).

This can probably be done by refactoring [`presc.misclassifications.misclass_rate.misclass_rate_feature`](https://github.com/mozilla/PRESC/blob/master/presc/misclassifications/misclass_rate.py#L5) to accept a metric function.

Issue #189, opened 2020-09-10
Refactor feature binning logic to separate utility function


dzeber:
We expect to use the binning methodology from [`presc.misclassifications.misclass_rate.misclass_rate_feature`](https://github.com/mozilla/PRESC/blob/master/presc/misclassifications/misclass_rate.py#L5) in other places. This should be moved out into a separate tested utility function.

PR #222, opened 2020-10-06
upload the ML_workfow

xinrongzhou:
upload the ML_workfow and rough implementation of #188


condition metrics(#188).ipynb
```diff
@@ -0,0 +1,1115 @@
...
+    "        \n",
+    "        accuracy_score_list=[]\n",
+    "        recall_score_list = []\n",
+    "        f1_score_list = []\n",
+    "        \n",
+    "        \n",
+    "        ThisGroup = []\n",
+    "        \n",
+    "        for index in range(1,len(total_histogram_counts)+1):\n",
+    "                ThisGroup = test_dataset[test_dataset['quantiles_'+feature]== str(index)]\n",
```
dzeber, line 882: Since you are only using the `qcut` result to group the rows, you don't need to add it to the dataset and worry about naming the column - you can just keep it as a variable (eg. just `x = pd.qcut(...)` above). Since it's created based on a DataFrame column, it'll have the same index and grouping will work as expected.

dzeber, line 882: Also, rather than iterating manually through the groups, you can simplifiy this by using `groupby`.

dzeber (changes requested):
Nice work! The ML workflow notebook is clear and you've taken the analysis to a good level of depth for the task.

I'm requesting a couple of admin changes:
- Please move the conditional metrics work to a separate PR (ie. move the commit that added that notebook to a new branch and open a new PR)
- Please add a row to the table in `datasets/README.md` for your dataset
- Notebooks should go in the `examples` dir. Let's make a new dir `examples/data_exploration` and put the ML workflow notebook in there. We can organize better once we get more notebooks added.

I really like what you've done in the ML workflow notebook for this task. I have a couple of questions and comments, and I put a couple of comments next to the code itself.
- Is there any particular reason you decided to restrict to those 3 columns of the dataset? I've seen other work that uses all the columns to predict `complication`.
- Nice use of SMOTE to adjust for the imbalance that shows up as the first model just predicts 0 for everything. This is a good example where the accuracy by itself is decent (87%) but hides the fact that the model is useless.
- It would be really great to add some markdown cells where you describe your observations and thought process, eg. regarding approach, model performance and data imbalance.
- I don't think the variable selection using forward regression is very helpful here, since you've only got 2 predictors, and the model it's computing importance for (OLS) is different from the SVM you're using. It's probably a good idea to keep the `forward_reference()` function definition for reference, but I would move it to a cell at the bottom.
- Nice to see the misclass_rate approach applied to this dataset! What do you make of those spikes in the plot for BMI?

For the conditional metrics, you should be writing your code in a `.py` file in the `presc` folder, but it's a good idea to import it into a notebook to test it. Since you are refactoring `misclass_rate`, you should go in and directly edit the code in that file!

Rather than computing results for a fixed list of metrics, I think we should pass the metric (ie. a function with an appropriate signature) in as an argument to the `metrics_feature` function.  Don't worry about adding a computation for the confusion matrix here, but you can use that as one of the examples when you try running this in a notebook.

Also, when you continue working on the binning logic as you mentioned at the bottom, you can split that work off in a separate PR as a solution for #189.

xinrongzhou:
- I have moved the ML_workflow notebook to new dir examples/data_exploration.

- I am working on modifying the ML_workflow based on your advice

ML_workflow_try.ipynb
```diff
@@ -0,0 +1,1450 @@
...
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "from sklearn.metrics import roc_curve\n",
+    "from sklearn.metrics import roc_auc_score\n",
+    "def plot_roc_cur(probs, y_test):\n",
+    "    \n",
+    "    \"\"\"\n",
+    "    Inputs: probs, precision, recall, thresholds = precision_recall_curve(y_true, y_scores)\n",
+    "#print(precision)\n",
```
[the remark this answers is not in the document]
xinrongzhou: Have deleted those lines. Sorry about that, they are just useless comments that I forgot to delete.

ML_workflow_try.ipynb
```diff
@@ -0,0 +1,1450 @@
...
+    "plot_roc_cur(probs, ytest)"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": 27,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "from imblearn.over_sampling import SMOTE\n",
```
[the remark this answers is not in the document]
xinrongzhou: Have added that.

ML_workflow_try.ipynb
```diff
@@ -0,0 +1,1450 @@
...
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": 29,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "# use the misclass rate feature to analysis this model\n",
+    "def misclass_rate_feature(\n",
```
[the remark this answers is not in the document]
xinrongzhou: Sure. Since I was editing the misclass_rate function while implementing it on this dataset, I left the function body there. Now I have deleted the body.

xinrongzhou:
Hi, David. Thank you so much for your comments.

- It is a piece of good advice to use complication as a prediction. I was thinking that maybe I can predict diabetes based on BMI and age, but this dataset is used to predict complications and now I have changed to that. 

- I also find SMOTE useful for this case. Even though I changed the prediction to complications, SMOTE still helps a lot.

- I have added some markdowns.

- the forward regression does not help in this case indeed. I will move it to the end as a reference.

- I have added a markdown that explains spikes in the graph.

condition metrics(#188).ipynb
xinrongzhou: Got that. I will make changes to that.

xinrongzhou: Got that.

xinrongzhou:
I have modified the implementation of issue#188 and put that in a new PR.

dzeber (approved):

dzeber:
Looks good, thanks!

Issue #208, opened 2020-09-10
Dataset wrapper API


dzeber:
For convenience we should wrap test datasets in an object that offers a common API, eg. accessing feature columns and label column, accessing pre-split train/test sets, etc. That way we don't need to pass around information about column names or indexing separately. There should also be methods for reading and writing from files.

We can probably create instances of these objects directly as a part of the dataset setup script #207 and work with these from then on.

PR #223, opened 2020-10-10
Dataset Wrapper API implementation and tests

AYYYang:
In response to issue #208. Implemented methods that allow access to raw dataset, split train and test by a ratio, getters for train/test dataset and features/label for raw dataset or train/test dataset.

Implemented tests for the implementation. Test coverage is 100%.


datasets/dataset_wrapper.py
```diff
@@ -0,0 +1,73 @@
...
+        Args:
+            test_dataset_name (string): a string that is the name of one of the test/datasets
+        """
+        
+        try:
+            # read the dataset from file path
+            self._dataset = pd.read_csv(dataset_file)
+
+            # set X,y
+            self.X,self.y = self._dataset.iloc[:,:-1], self._dataset.iloc[:,-1]
```
dzeber, line 30: I think we should pass in the label column name as a string to the function, and use that to identify the X and y parts. That way we don't need to make assumptions about column positions. We will want to use this class to handle all datasets interacting with PRESC, so we should stay general.

datasets/dataset_wrapper.py
```diff
@@ -0,0 +1,73 @@
...
+            raise
+
+    def split_test_train(self,test_size:float = 0.2, random_state:int = 0) -> None:
+        self.X_train, self.X_test, self.y_train, self.y_test = train_test_split(self.X, self.y, 
+        test_size=test_size, random_state=random_state)
+
+    def get_raw_dataset(self) -> object:
+        return self._dataset
+
+    def get_label(self, subset:str = None) -> object:
```
dzeber, line 44: Would be good to add a docstring here and below to document the `subset` parameter.

tests/test_dataset_wrapper.py
```diff
@@ -0,0 +1,69 @@
+import numpy as np
+import pandas as pd
+import pytest
+from sklearn.model_selection import train_test_split
+
+from datasets.dataset_wrapper import DatasetWrapper
+
+wine_data_path = "../datasets/winequality.csv"
```
dzeber, line 8: Hardcoding the path here is not a great idea as it makes assumptions about which working dir the tests are run from.

In fact I noticed this because I tried running the tests from the root dir :).

A way to get around that is to use
```python
from pathlib import Path
Path(__file__)
```
to get the path of this module (`test_dataset_wrapper.py`), and set paths relative to that.

tests/test_dataset_wrapper.py
```diff
@@ -0,0 +1,69 @@
...
+random_state, test_size = 0, 0.2
+
+# test invalid initailization of DatasetWrapper object
+def test_fie_does_not_exist():
+    with pytest.raises(FileNotFoundError):
+        DatasetWrapper("datasets/winequalit.csv")
```
dzeber, line 14: I think this and the next test assumes a different working dir than `wine_path_data` above?

tests/test_dataset_wrapper.py
```diff
@@ -0,0 +1,69 @@
...
+
+def test_fie_format_incorrect():
+    with pytest.raises(IOError):
+        DatasetWrapper("datasets/README.md")
```
dzeber, line 18: I think this is testing the failure when you pass in a file that is not a CSV?

For me, this is actually __not__ failing, because Pandas is trying to be smart and is parsing the table we have in that README. If you're not getting this behaviour, please make sure you've activated the conda environment before running the tests.

On the other hand, I tried running this with a different file and it's giving a Pandas `ParserError` rather than `IOError`.

dzeber (changes requested):
Great work on this PR! The API is nice and clean, and the tests do a good job of covering the functionality.

I'm requesting a couple of updates before merging. Top-line:
- please move `dataset_wrapper.py` to the `presc` dir - all core code should be in there.
- make sure to run Black formatting prior to submitting. This should happen automatically if you set up `pre-commit` as described in the [README](https://github.com/mozilla/PRESC#getting-started). When you run `git commit`, it will automatically run Black, and if Black made changes to the files, the commit will fail and you will need to `git add` those changes.

There are a couple of other minor things I've noted in inline comments.

datasets/dataset_wrapper.py
AYYYang: Changed the function to include a label parameter. Also added a label setter function.

AYYYang: Added a `label` parameter for this function and added a label setter function as well.

datasets/dataset_wrapper.py
AYYYang: Added a docstring in the `get_label()` function.

tests/test_dataset_wrapper.py
AYYYang: Took your suggestion and now everything is in reference to Path! Thank you for the catch.

tests/test_dataset_wrapper.py
AYYYang: Right. I changed this to using the reference path as well.

tests/test_dataset_wrapper.py
AYYYang: I passed in a python script just to make this test fail. Not sure if we need to handle files like markdown as errors? I

AYYYang:
Also moved the wrapper class to the presc folder. `pre-commit` is now installed and running before I commit. Thank you for the comments David!

tests/test_dataset_wrapper.py
dzeber: Using the setup.py script is fine. Anything that is not CSV should raise the error. What I meant was the markdown should have raised the exception as well, and I think that's what you intended, but it didn't because of a Pandas feature trying to recognize that the markdown in fact contains a table.

dzeber (approved):

dzeber:
Looks great! Thanks for making the fixes.

PR #224, opened 2020-10-13
raw implementation of condition metrics(#188)

xinrongzhou:
- This is just a new PR with the same old conditional metrics notebook.

- I will keep working on the details and I will put my modified work in the misclass_rate.py.


xinrongzhou:
I have deleted the notebook I worked on for the implementation. I have put my updates in the file misclass_rate.py. I am so sorry for the misunderstanding before. I am looking forward to your feedback of my work! Thank you so much for your patience.

presc/misclassifications/misclass_rate.py
```diff
@@ -123,8 +154,51 @@ def misclass_rate_feature(
...
+    
+    
+    # compute metrics
+    
+    metric_list=[]
+    
+    if metric == "accuracy score":
+        accuracy_score_list=[]
+
+        for index in range(1,len(total_histogram_counts)+1):
```
dzeber, line 166: It would be better not to keep repeating this code block. Looks like the only thing that changes is the metric function? This can be saved to a variable above, eg. `metric_fun = accuracy_score`, and then use `sc = metric_fun(test_y, pred_y)`. In fact, if the `metric` param accepts a function as discussed above, you could just use that function directly (`metric(test_y, pred_y)`).

dzeber (commented):
Sorry for the long delay in getting you feedback!

This is coming along well. Please be sure to update `tests/test_misclassifications.py` to test the new functionality and make sure the tests pass.

Also, please run Black formatting prior to submitting. This should happen automatically if you set up `pre-commit` as described in the README. When you run `git commit`, it will automatically run Black, and if Black made changes to the files, the commit will fail and you will need to `git add` those changes.

presc/misclassifications/misclass_rate.py
```diff
@@ -123,8 +154,51 @@ def misclass_rate_feature(
...
+                    theGroup = groups.get_group(str(index))
+                    test_y = pd.Series(theGroup[y_test_name]).to_numpy()
+                    pred_y = pd.Series(theGroup['pred']).to_numpy()
+                    f1_sc = f1_score(test_y,pred_y)
+                    f1_score_list.append(f1_sc)
+        metric_list = f1_score_list
+                
+    del test_dataset['pred']
+
+    return bins, misclass_rate_histogram, standard_deviation,,metric,metric_list
```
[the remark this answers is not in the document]
xinrongzhou: have deleted that.

presc/misclassifications/misclass_rate.py
```diff
@@ -9,6 +9,7 @@ def misclass_rate_feature(
     categorical=False,
     bins=10,
     bins_type="regular",
+    metric = "accuracy score"
```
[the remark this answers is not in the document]
xinrongzhou: Good idea. Done that.

presc/misclassifications/misclass_rate.py
xinrongzhou: It is a really good idea. I have modified it.

dzeber:
Hi @xinrongzhou, please be sure to remove draft status when this is ready for review.

xinrongzhou:
I have updated the misclass_rate.py and the tests. It is ready for review now.

xinrongzhou:
If there is something more I need to do, please let me know.

presc/misclassifications/misclass_rate.py
```diff
@@ -50,25 +51,52 @@ def misclass_rate_feature(
...
+            groups = test_dataset.groupby(
+                by=pd.qcut(test_dataset[feature], q=bin_num, duplicates="drop")
+            )
+
+        # groups for regular bin edges or known bin edges
+        else:
+            groups = test_dataset.groupby(by=pd.cut(test_dataset[feature], bins=bins))
 
         # Histogram of all points
         total_histogram_counts, bins = np.histogram(test_dataset[feature], bins)
```
dzeber, line 102: You can now remove the computations involving `total_histogram_counts` and `misclass_histogram_counts`, since the function now computes more general metrics than just misclassification rates.

presc/misclassifications/misclass_rate.py
```diff
@@ -78,7 +106,36 @@ def misclass_rate_feature(
...
+            theGroup = groups.get_group(key)
+            test_y = pd.Series(theGroup[y_test_name]).to_numpy()
+            pred_y = []
+            for row in theGroup.index:
+                pred_y.append(y_pred_series.at[row])
+            sc = accuracy_score(test_y, pred_y)
+            metric_list.append(sc)
+
         # Histogram of all points for categorical features
         total_histogram_counts = test_dataset[feature].value_counts().sort_index()
```
dzeber, line 140: Similarly as  above, you can now move the rest of this outside of this function.

dzeber (commented):
Looks like the computation and tests are done. I would like to do some final refactoring and then we can merge.

I've dropped some suggestions for reducing duplicated code.

Also, `misclass_rate_feature` now no longer just computes the misclassification rate, as you have generalized it to compute any metric. This should be reflected by changing the name of the function and updating the documentation.

That said, we should keep the original functionality of `misclass_rate_feature` for compatibility as well, but it should use your new code to do the computation. This also helps test your new code, as the original tests for `misclass_rate_feature` should still pass using this new version.

 Here's my suggestion for how to achieve this:

- rename the current `misclass_rate_feature` function to something more appropriate, eg. `compute_conditional_metric`
- create a new function named `misclass_rate_feature` with the original signature and return values (ie as it was prior to your PR). It should produce the same result as the original function, but call the new `compute_conditional_metric` to do the computation.
- As a part of this refactoring, you would remove code from `compute_conditional_metric` that is no longer relevant, eg the computations for `total_histogram_counts`, `misclass_histogram_counts`, `standard_deviation`, etc and move it to the new `misclass_rate_feature`, updating as necessary.  For example, you could use a `metric_function` that returns 2 values: the size of the group (corresponds to `total_histogram_counts`) and the number of misclassifications (corresponds to `misclass_histogram_counts`).

xinrongzhou:
Thank you so much, David! Great advice! I will work on them right now.

xinrongzhou:
- Do you mean that I use two different functions: `misclass_rate_feature` and `compute_conditional_metric`. In `compute_conditional_metric`, I only do the computation. In `misclass_rate_feature`, I keep it as old? And in `misclass_rate_feature`, I call `compute_conditional_metric`to do the work?  

- How about returns and parameters?  For the new functionality, I add a parameter `lambda function...` and a return value `metric _list`to the original `misclass_rate_feature`, Shall I leave them in `misclass_rate_feature`?

- I am kind of confused with 
> For example, you could use a metric_function that returns 2 values: the size of the group (corresponds to total_histogram_counts) and the number of misclassifications (corresponds to misclass_histogram_counts).

 What is this `metric_function`? Is this a new function that I call in `compute_conditional_metric`? 

Thank you so much for your help!

dzeber:
>  Do you mean that I use two different functions: `misclass_rate_feature` and `compute_conditional_metric`. In `compute_conditional_metric`, I only do the computation. In `misclass_rate_feature`, I keep it as old? And in `misclass_rate_feature`, I call `compute_conditional_metric`to do the work?

Yes, exactly. Currently you've modified the `misclass_rate_feature` to fill `metric_list` with the computation over the groups, which is good. However, it also still includes the old computation for `misclass_rate_histogram` (eg [here](https://github.com/mozilla/PRESC/blob/37bec204f03480bd421b35716172fcfdab137690/presc/misclassifications/misclass_rate.py#L140)). What I'm suggesting is:

- rename the current `misclass_rate_feature` to `compute_conditional_metric`
- create a new function `misclass_rate_feature` with the original signature that calls `compute_conditional_metric` and gives the same result as the original function.

So you would have something like:
```python
def compute_conditional_metric(...):
    # your new/modified code:
    # grouping
    # for loop over groups
    metric_list.append(...)
    return bins, metric_list

def misclass_rate_feature(
        test_dataset,
        test_predictions,
        feature,
        categorical=False,
        bins=10,
        bins_type="regular"):
    # metric_func computes [total count in each group, misclass count in each group]
    metric_func = lambda y_true, y_pred: [len(y_true), (y_true != y_pred).sum()]
    bins, metric_list = compute_conditional_metric(..., metric_function=metric_func)
    misclass_rate_histogram = ... # divide metric_list[i][1] / metric_list[i][0]
    # copy in the standard deviation calculation
    return bins, misclass_rate_histogram, standard_deviation
```

So then, the original tests for `misclass_rate_feature` should still run, since you've only changed the inner computation (this is a good check that the new code is consistent with the previous behaviour).

>   How about returns and parameters?  For the new functionality, I add a parameter `lambda function...` and a return value `metric _list`to the original `misclass_rate_feature`, Shall I leave them in `misclass_rate_feature`?

As in the snippet above, `compute_conditional_metric` would only need to return `bins, metric_list`, and `misclass_rate_feature` should return `bins, misclass_rate_histogram, standard_deviation` (same as it did prior to your PR).
 
> What is this `metric_function`? Is this a new function that I call in `compute_conditional_metric`?

As in the snippet above, whenever you use `compute_conditional_metric` you would need to provide a `metric_function`. Because in the original misclass rate you need both the totals and the misclass_counts, I suggest using a `metric_function` that just returns these two values for each group.

I hope this helps - happy to discuss further on Slack.

xinrongzhou:
Cool! I get that.

presc/misclassifications/misclass_rate.py
```diff
@@ -78,7 +106,36 @@ def misclass_rate_feature(
...
 
+        # compute metric
+        metric_list = []
+        # get all groups and compute their metrics for each group
+        for key in groups.groups.keys():
+            try:
+                theGroup = groups.get_group(key)
+                test_y = pd.Series(theGroup[y_test_name]).to_numpy()
+                pred_y = []
+                for row in theGroup.index:
```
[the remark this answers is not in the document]
xinrongzhou: Good idea. Have modified that.

presc/misclassifications/misclass_rate.py
xinrongzhou: Removed that.

presc/misclassifications/misclass_rate.py
xinrongzhou: I kept `total_histogram_counts = test_dataset[feature].value_counts().sort_index()`, since I want it to help me calculate bins after that:` bins = np.asarray(total_histogram_counts.index)`. This is the easiest way I can think to calculate bins for categorical data. Can you give me some advice about that?

xinrongzhou:
Hi, David! 
I have modified the `misclass_rate.py`: I have added `compute_conditional_metric` function and modify the `misclass_rate_feature` as you suggested. This new version passes all the original tests in `test_misclassifications.py`(prior to my PR). Could you please take a look at the updated `misclass_rate.py` and make sure I am following your advice correctly. Thank you so much!
Also, I would like to discuss that:

- I had written tests that test the new functionality together with the old misclass_rate_feature( I added new parameters and new return values). Since I had put new functionality inside the old `misclass_rate_feature` function. Now since I got a new function `compute_conditional_metric`, shall I write a new test function named`test_compute_conditional_metric` and test new functionalities here?

- Since the metric was in `misclass_rate_feature`, I plotted the `misclass_rate` and `metric_list` together in the `show_misclass_rate_feature`( I plotted two figures for each feature). Now how should I plot the conditional metric figure? 
I was thinking: you said the `compute_conditional_metric` is more general than `misclass_rate_feature`. Maybe the show function can be generalized to show any kind of `conditional_metric` instead of simply a `misclass_rate`. This is just my rough idea, can you give me some advice about that?

Thank you so so much for your help. If there is something more I need to do, please let me know.

presc/misclassifications/misclass_rate.py
dzeber: Since the bins are already given by the `groups` you could just use `list(groups.groups.keys())`.

dzeber:
> Could you please take a look at the updated misclass_rate.py 

Yes, this looks great! Nice work 🙂 

> shall I write a new test function namedtest_compute_conditional_metric and test new functionalities here?

Yes, we should have a test for this function. For now, maybe just add one test function very similar to `test1_misclass_rate_feature`, ie. using the same inputs and a `metric_function` (eg. accuracy).

> Now how should I plot the conditional metric figure?

You can use the same plot. It's essentially just plotting `metric_list` vs `bins` (but with different variable names), so that will work for the more general case as well.

xinrongzhou:
Hi, David! I have finished the tests and comments. Please take a look and let me know if there is anything further to do.

dzeber:
Thanks for this! It's good to go.

PR #226, opened 2020-10-25
Added the kick starter data exploration ML workflow

AYYYang:
1. Updated the README.md file
2. Left the dataset not pushed, as the size is too large


dzeber (changes requested):
This is really great work! The data cleaning steps and preprocessing show careful attention to detail, and I like that you tried out a couple of different models and included parameter tuning.

I'm requesting a couple of minor updates:

- Please update the `environment.yml` file to include the new packages, so that the notebook is runnable from the environment.
- Make sure the notebook runs as is. I think the numerical and feature columns need to be updated in cell 16.

Some general comments (not "required" as changes for this PR unless you'd like to):

- Is the reason you are converting the sparse matrix back to a DataFrame after preprocessing in order to handle missing values? The scikit-learn models and functions like `train_test_split` do access sparse matrices. Either way, it would be good to add a comment about this.
- Seeing as `category` has a lot (>150) of values, it would be worthwhile checking how much predictive power it contributes overall, and if we can get by without it. The downside is that the one-hot encoding introduces so many new columns, which can affect the stability of some models. On the other hand, the dataset is large enough to support it. You don't need to change anything for this PR - just wanted to call this out as a consideration.
- Best practice is to fit the transformers on the training data only - otherwise the transformations (eg. mean & stdev) are computed including the test data as well. This can be handled using the scikit-learn Pipeline (see the grey box at the end of [this section](https://scikit-learn.org/stable/modules/cross_validation.html#computing-cross-validated-metrics) in the guide for an example)
- Interesting the biggest issue in the confusion matrix is that the classifier is having trouble telling `cancelled` apart from `failed`, which makes sense as they both have funding falling short of the goal. This will be a good test case for PRESC - to better understand why this is the case :) 
- The data cleaning at the beginning is pretty well documented. It would be great to add a couple more markdown cells with your observations and interpretations.

AYYYang:
Thank you David for the thorough feedback!
I added pandas profiler to the environment.yml file, and re ran the notebook before I committed.
I also changed the notebook to use pipeline and added code/comments/markdown cells in regard to the comments you gave. 

:) Thanks again, I really appreciate the feedback!

dzeber (approved):

Issue #193, opened 2020-09-10
Implementation for conditional feature distributions


dzeber:
Implement the computation for [conditional feature distributions](https://github.com/mozilla/PRESC/blob/master/docs/ROADMAP.md#conditional-feature-distributions).

alberginia:
I implemented this one for my Outreachy application:
https://github.com/mozilla/PRESC-Outreachy-archive/blob/master/dev/alberginia/alberginia_issue7_3datasets.ipynb

AYYYang:
@dzeber @alberginia 
I think I understand most of the specification, however, I have a question regarding the input.
Two of the allowed inputs are: `predicted scores or a function of the features`
I am assuming that `predicted score` means the `y` column in the test dataset. I am not sure what should be the output, more specifically, let's say if we are plotting a histogram, what should be on the x-axis?

For `function of the features`, I am not sure what this means.

dzeber:
Sorry for the delay in getting you feedback on this!

The goal would be, given a column of data, to create a separate histogram corresponding to the subset of rows that fall into each cell of the confusion matrix. For example, with two classes there would be 4 histograms. To determine this grouping, you can use the column of true labels (eg. `y_true`) and the column of predicted labels (eg. `y_pred`) to get groups eg. (eg. `y_true == 1 and y_true == y_pred`, `y_true == 1 and y_true != y_pred`, `y_true == 2 and y_true == y_pred`, `y_true == 2 and y_true != y_pred`). Note that we don't need to compute the counts as in the actual confusion matrix, but rather we are just labelling each data row according to which cell it would be counted in, in order to group on this label.

Re your question about the input, this "column of data" could in principle be any feature in the original dataset or any computed column, eg. the average of two features or the scores predicted by the model. The code should be the same either way, since it should work the same so long as the input is a pandas series or list the same length as the dataset. So you wouldn't need to do anything special to support these different cases, but they should "just work" when testing in a notebook.

Re the output, it would always be a histogram, so the output would have bins over the range of column values on the x-axis and (relative) frequency counts on the y-axis. (In the case of categorical data, it will be a barplot instead, but it follows the same idea).

I think the outstanding work that needs to be done for this issue is:
- copy the existing code into the presc package dir
- implement splitting according to the confusion matrix cells as described above
- maybe refactor the existing code to have a separate function that generates a single histogram, and call that inside the `for` loop in `histograms()` to avoid code duplication
- add a couple of tests
[...]

PR #227, opened 2020-10-25
Implementation of conditional feature distribution

AYYYang:
Basically made a function to get confusion matrix based on the test labels and predicted labels inputs. Not sure if this is the functionality that's asked for.

One major question is that whether I should just re use the code @alberginia wrote for the histogram and 2D projection and change the function parameters, or re write the functions. If former, should I just make a helper file that contains the code contributed by @alberginia?


dzeber:
Sorry for the delay on getting you an answer on this. I will update tomorrow.

dzeber:
I've posted some more details in the issue (#193). Yes, I think you should continue as you have started here by migrating the existing code to the new file you created. You shouldn't need to fully rewrite the function, but I encourage you to refactor if it would simplify the code or reduce code duplication.

Let's leave the 2D projection approach for a separate issue, but that would definitely be a good addition as well.

AYYYang:
This commit is definitely buggy. But I think the general logic is there. Will write tests and debug later today.

AYYYang:
The basic functionality should be working. See the notebook file for reference.

By basic functionality I meant:
1. The function `plot_all_histogram_conditional_feature_distribution()` should plot histograms corresponding to each quadrant on the confusion matrix plot as **individual plots**, as long as there is a >0 count for that quadrant. This should work for any of the test datasets we have in our `datasets/` folder.

2. The function also automatically calculates a bin number using the [Freedman Diaconis rule](https://en.wikipedia.org/wiki/Freedman%E2%80%93Diaconis_rule#:~:text=Freedman%20and%20Persi%20Diaconis.,under%20the%20theoretical%20probability%20distribution.)

Some TODOs  I am thinking about:
1. Also allow the function to generate bar plots for categorical data. For now, we do not have any test datasets in the `datasets/` folder. Should I use other data for testing?
2. I thought about putting all the histograms on the same page, and order them the same way as the `plot_confsion_matrix_()` from the sklearn module. However, sometimes we have groups/quadrants that have 0 data. One way to address this is to pass in all the possible group labels from the raw data. Should I integrate this? 

Thanks in advance!

AYYYang:
The function now should work for categorical data as well. I tried the function with trivial categorical data. Added tests for the function.

Let me know what you think! It has been really interesting so far!

dzeber (commented):
Thanks, this is really great work! Well-organized code with good test coverage. I have a couple of pretty minor comments, but otherwise this is done.

The main thing we need to change is to use `block=False` in `plt.show`, because otherwise this blocks the tests from running.

Aside from this, what you have works fine but I've made a couple of suggestions to streamline things using built-in pandas functions (easier to read and maintain).

Also, please move the notebook to the `examples` dir.

dzeber:
> The basic functionality should be working.

Yes, it works just as described in the roadmap :)

>  1. Also allow the function to generate bar plots for categorical data. For now, we do not have any test datasets in the `datasets/` folder. Should I use other data for testing?

What you have using the synthetic categorical data is fine. Once Juan's PR gets merged, that will bring a fully categorical dataset, so we can use that.
 
>  2. I thought about putting all the histograms on the same page, and order them the same way as the `plot_confsion_matrix_()` from the sklearn module. However, sometimes we have groups/quadrants that have 0 data. One way to address this is to pass in all the possible group labels from the raw data. Should I integrate this?

I like that idea for presentation, but let's leave that until we get more concrete about the layout of the report.

AYYYang:
> > The basic functionality should be working.
> 
> Yes, it works just as described in the roadmap :)
> 
> > 1. Also allow the function to generate bar plots for categorical data. For now, we do not have any test datasets in the `datasets/` folder. Should I use other data for testing?
> 
> What you have using the synthetic categorical data is fine. Once Juan's PR gets merged, that will bring a fully categorical dataset, so we can use that.
> 
I took the suggestions you gave and thank you again for being so helpful! Regarding Juan's PR, I saw that it is merged now. Should I just pull all the work he did, ie pull all the new changes in master to my local branch and continue working on it? Or should I open a new branch based on master once this PR is merged? 
> > 1. I thought about putting all the histograms on the same page, and order them the same way as the `plot_confsion_matrix_()` from the sklearn module. However, sometimes we have groups/quadrants that have 0 data. One way to address this is to pass in all the possible group labels from the raw data. Should I integrate this?
> 
> I like that idea for presentation, but let's leave that until we get more concrete about the layout of the report.
Sounds good. I will continue working on this once that is decided.

dzeber:
> Should I just pull all the work he did, ie pull all the new changes in master to my local branch and continue working on it? Or should I open a new branch based on master once this PR is merged?

As we discussed, now that these have been merged, you would fetch `upstream` (main repo), merge into your local master (should just be a fast-forward), then `push origin master` (your GH fork). It's best to then start a new branch off your updated master, and you can delete the local branch you used for this PR.

PR #228, opened 2020-11-10
Categorical data ML workflow

xinrongzhou:


dzeber:
Thanks for contributing this! This dataset is a good test case. It is quite small (n=66), meaning that we need to be careful about not splitting off too small a test set - and in fact, at this size we could use leave-one-out CV instead. Also, seems like the logistic model is doing pretty badly (predictions like random chance). I haven't looked into it too much, but would be nice to try to diagnose using the PRESC tools!

PR #229, opened 2020-11-17
PR for testing 

jdiego-miyashiro:


dzeber (commented):
This is a good start. Nice use of the fixture to set up the `SpatialDistribution` instance. A couple of comments:

- It's usually best to make tests as deterministic as possible. I would set up a fixed DataFrame in `catd_analysis`, and use a fixed choice of row to compare in the tests.
- It's fine to make the test data really small, since you just want to make sure that the code does what is expected. For this case, you might want to use ~8-10 rows, 2-3 columns with ~2-3 unique values per column - just enough to make the computations non-trivial.
- For the distance metric tests, you should actually check you get the result you expect (from the small deterministic dataset). So you would `assert catd_analysis.<dist>(dpoint1, dpoint2) ==` whatever you get doing the calculation by hand.
- Comparing floats with `==` will sometimes fail due to rounding. This can be handled using [`numpy.testing.assert_approx_equal`](https://numpy.org/doc/stable/reference/generated/numpy.testing.assert_approx_equal.html) instead.

Also, please rebase onto the latest master so that this PR doesn't include previous commits that were already merged:

```bash
git fetch upstream  # pull latest updates from main repo
git checkout master  # make sure you're on your local master
git merge upstream/master  # merge the upstream changes into your local master
git push origin master  # sync to your GH fork master
git checkout PR-test-suit  # check out the branch for this PR
git rebase master  # apply your new commits on top of updated master.
git push origin PR-test-suit  # update PR
```

Your PR should now only show the new commits after the previous PR got merged.

dzeber:
This is much improved - great job! There is one last thing: please use `block=False` inside the `plt.show()` calls in `categorical_distance.py`. This will stop the plots showing up and blocking execution when the tests are run.

PR #230, opened 2020-11-19
Evaluation implementation outline

dzeber:
This lays out an API for working with datasets and models and an implementation design for evaluation methodologies.

Summary of changes:

- Update the DatasetWrapper API and rename to Dataset
- New wrapper for classification model as used in PRESC evaluations
- Sample evaluation implementation in `_example_misclass_rate.py`
  * this ports `misclass_rate_feature()`, `show_misclass_rate_feature()`, `show_misclass_rates_features()` from `misclass_rate.py`
  * same functionality: code is copied over and refactored to use new API
  * config options managed at the instance level
- Sample notebook `_evaluation_example.ipynb` demonstrating usage

The design here is subject to change. In particular, handling of config options should look more or less as it is here, but we wil likely replace the implementation with something slightly more sophisticated.

I will follow up in a separate PR with an initial report template.

Flagging @AYYYang for review as I have refactored the DatasetWrapper class. In particular, loading from file is left outside of the class as I think the main use case will be for a preloaded DF (we don't need to handle filetypes etc). Also, properties are made accessible as attributes flattening the train/test structure.


AYYYang:
Thank you David! It was a good learning opportunity to see the refactored code!

PR #232, opened 2020-11-29
PR extension to numerical data

jdiego-miyashiro:
Pull request that implements the extension of categorical_distance.py to numeric values


presc/spatial_distribution/categorical_distance.py
```diff
@@ -199,113 +247,182 @@ def goodall_frequency(self, counts):
...
                               estimate the average distance and if greater indicates the precise number
-                              of samples to be take"""
+                              of samples to be take
+        Returns(float): Estimate of average distance from a point to the data"""
 
-        if distance_sample < 1:
+        metric, type = self.get_metric(metric)
+        if type == "categorical":
+            data = self.cat_data
+            dpoint = dpoint[dpoint.apply(lambda x: isinstance(x, str))]
```
dzeber, line 264: To subset the categorical columns (and numerical columns below), it would be nicer to just select the corresponding elements of the datapoint vector directly. `dpoint` should be a Series indexed by the column names, so you should be able to do `dpoint[data.columns]`

dzeber (commented):
This is looking good. Please update the tests to cover the new code (eg. add a column or two of numerical data to the toy dataset and run the numerical case on the computations).

For a dataset with both discrete and continuous columns, how can we compute distances taking all columns into account? Currently it looks like it does either one or the other, depending on the metric. It would be good to compute an overall metric that combines the two. There are different ways to do this, but to start simple we could just add together the distances for categorical columns and for numerical columns (ie. this could be handled in `distance_to_data`). To avoid issues with scaling, we probably want to standardize the numerical data, eg using [StandardScaler](https://scikit-learn.org/stable/modules/generated/sklearn.preprocessing.StandardScaler.html#sklearn.preprocessing.StandardScaler).

PR #234, opened 2020-12-04
Added an example of the conditional feature distribution notebook

AYYYang:
David please let me know if you think it is worthwhile to spend time on this:

1.  Fix the labels of the figures, currently they ugly underlines.
2. Set the white space distance/edgecolor? to be consistent with the example miss-classification rate output figures.


dzeber:
Thanks - this looks good. The one comment I have is that you don't need to actually train the model in this notebook. The [`cm` object](https://github.com/mozilla/PRESC/blob/master/presc/model.py) has the test set predictions precomputed, so you can just do `y_predict=cm.test_predictions`.

To get this to show up in the report, you also need to add an entry in the [TOC file](https://github.com/mozilla/PRESC/blob/master/presc/report/_toc.yml) under `chapters:`. Add this and see if you're able to run it locally following the steps [here](https://github.com/mozilla/PRESC/blob/master/docs/REPORT.md).

> please let me know if you think it is worthwhile to spend time on this:

Not at this point, as we will be migrating to a different visualization library (probably starting in Jan). If you have more coding time left, the best thing would be to work on refactoring your feature to a class under in the `evaluations` dir (following the example of [evaluations/_example_misclass_rate.py](https://github.com/mozilla/PRESC/blob/master/presc/evaluations/_example_misclass_rate.py).

AYYYang:
Thanks David! Did modification and will work on it some time maybe after my finals!

PR #245, opened 2021-02-22
Closes #238 Add config file for management of config option

### LINKED ISSUE #238: Management for config options
Many components of the PRESC computations should be customizable by a user running the tools in their own environment. Some options are currently passed around as function args, but we need a more unified way of setting config options that doesn't require code changes, eg. via a config file.

Some of the components that will be configured via options are:

- The evaluation computations, eg. the binning scheme to use for grouping test samples
- Plot components, eg. element sizing or plot arrangement
- How an evaluation applies to a dataset, eg. some would be run per-feature, and we'd want to include or exclude certain features. Also, some of the per-evaluation options may be allowed to vary per-feature in this case.
- The report generation, eg. which evaluations to include

gleonard-m:
Closes #238 Add config file for management of config option

### DIFF
--- a/presc/evaluations/conditional_distribution.py
+++ b/presc/evaluations/conditional_distribution.py
@@ -4,8 +4,6 @@ from numpy import histogram, histogram_bin_edges
 from pandas import Series, MultiIndex
 import matplotlib.pyplot as plt
 
-DEFAULT_CONFIG = {"binning": "fd", "common_bins": True, "plot_width_fraction": 1.0}
-
 
 def _histogram_bin_labels(bin_edges):
     """Returns a Series of labels for histogram bins generated using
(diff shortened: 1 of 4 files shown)


dzeber (approved):
This looks great. It looks like the entrypoints where the config can be passed are now narrowed down to the evaluation object creation (eg. ConditionalMetric) and the `compute_` function, which makes more sense.

Do you think it would be worthwhile having a package level config context, eg. as a module that can be imported which loads the YAML file? I guess this depends on how the library is used. To run the report, each notebook would start its own kernel, so that would reread the config file each time (maybe less ideal). If the library is imported into a notebook and used interactively, however, that may help. What do you think?

dzeber:
Merging this now, and we can discuss that question later as it doesn't block the changes here.

PR #259, opened 2021-02-25
Improvements to report execution

### LINKED ISSUE #241: Add a unified entrypoint to run the report
We need a better interface to run the report from, and to remove any dependency on the [example dataset/model](https://github.com/mozilla/PRESC/blob/master/presc/report/setup_report.py) we are currently using in the report.

As a part of this, we need a way to share the required inputs (data/model/config) between the notebooks. Looks like the simplest approach would be to serialize and reload separately in each notebook.

dzeber:
Resolves #241. 

Summary of changes:

- The report can now be run using the Python API in `presc.report.runner`, taking model & dataset as inputs
- Report source moved into subdir `resources`
- Data is shared between notebooks using a data store (`presc.report.runner.Context`) on disk
- `setup_report.py` moved to `examples/`. This script gives an example of the setup code the user would need to set up the inputs and trigger the report.
- `add_tags.py` moved to `scripts/`. This no longer needs to be part of running the report. We will now assume that it has been run on the report notebook prior to committing changes.

@gleonard-m please take a look and see what you think of the approach. Tests are still to follow.


environment.yml
```diff
@@ -16,6 +16,7 @@ dependencies:
   - pre-commit=2.6.0
   - pytest=5.4.3
   - pytest-cov=2.10.0
+  - pyyaml=5.4.1
```
gleonard-m, line 19: Did I miss adding this in my PR for the config file?

dzeber: I think so - didn't see it in here. It's already in the environment as a dependency from the other libraries, but good to add list entries for the ones we import directly.

dzeber:
I've added tests covering the `presc.report.runner` module. This also has the effect of test-running the report on a small sample dataset with the default settings. I think this is ready to merge.

tests/test_runner.py
```diff
@@ -0,0 +1,304 @@
...
+        toc = [x.strip() for x in f.readlines()]
+        notebooks = [x[8:] for x in toc if x.startswith("- file: ")]
+    for nb in notebooks:
+        assert (rr.execution_path / f"{nb}.ipynb").exists()
+        assert (rr.report_main_page.parent / f"{nb}.html").exists()
+    # Error message shows in notebooks
+    with open(rr.report_main_page.parent / "conditional_metric.html") as f:
+        eval_html = f.read()
+    assert "Error" in eval_html
+    assert "Traceback" in eval_html
```
gleonard-m, line 304: Very thorough runner test coverage!

presc/report/runner.py
```diff
@@ -0,0 +1,296 @@
...
+        # path.
+        if self.report_main_page.exists():
+            # Symlink the main page to the top level for convenience.
+            try:
+                main_page_target = self.report_main_page.relative_to(
+                    self._linked_main_page.parent
+                )
+                self._linked_main_page.symlink_to(main_page_target)
+            except OSError:
+                pass
```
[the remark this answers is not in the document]
dzeber: The `report_html` accessor will give the symlink if it exists, otherwise the target report page. I thought that would be sufficient as the user would be working from Python anyway. This is not yet really documented but it will be 🙂. Not really sure how useful the symlink thing will be, I was just thinking to add some sugar to make it a little more user-friendly.

tests/test_runner.py
dzeber: :) good to cover our bases at the entrypoint

gleonard-m (approved):

PR #265, opened 2021-03-05
Closes #263: Expands and implements config options

### LINKED ISSUE #263: Organize & expand config options
- [ ] Develop an organization structure for the config options to be used in different parts of the library
- [ ] Implement basic report-level options controlling which notebook get included
- [ ] Implement basic evaluation-level options controlling the scope of evaluation execution, eg. which columns to include & per-column overrides

dzeber:
The config structure now supports controlling the pages showing in the report, the columns for which the evaluations are computed, and per-column overrides for the computation options.

The pattern I went with for config in the evaluations is:

- the `compute_` module function takes all options as explicit params and stores them as attributes in the Result instance
- the evaluation main class takes options from a config dictionary, but may allow for overrides via kwargs

The main class is the main entrypoint to running the evaluation as is on a given dataset, and the `compute_` top-level function exposes the underlying computation for exploration/development purposes.

The tests I added for the evaluations only check that the config settings get picked up. The actual computations are not yet tested (this is still an open issue).

### DIFF
--- a/presc/evaluations/conditional_distribution.py
+++ b/presc/evaluations/conditional_distribution.py
@@ -1,4 +1,5 @@
 from presc.evaluations.utils import is_discrete
+from presc.utils import include_exclude_list
 
 from numpy import histogram, histogram_bin_edges
 from pandas import Series, MultiIndex
@@ -16,7 +17,7 @@ def _histogram_bin_labels(bin_edges):
 
 
 def compute_conditional_distribution(
-    data_col, true_labs, pred_labs, as_categorical=False, config=None
+    data_col, true_labs, pred_labs, as_categorical=False, binning="fd", common_bins=True
 ):
     """Compute a distributional summary.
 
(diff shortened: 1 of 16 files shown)


presc/evaluations/conditional_distribution.py
```diff
@@ -138,46 +148,58 @@ class ConditionalDistribution:
...
-            across groups (`True`) or computed within each group (`False`), default: True
-        `plot_width_fraction`: width of the bars relative to available space on the plot.
-            Smaller means more space between the bars, default: 1.0
+    config: the main config dict
     """
 
-    def __init__(self, model, test_dataset, config=None):
-        self._config = config
+    def __init__(self, model, test_dataset, config):
+        self._config = config["evaluations"]["conditional_distribution"]
```
gleonard-m, line 155: Definitely prefer passing in the entire config and having the Class/Object/function pulling what it needs

gleonard-m (approved):
Nice improvement on making PRESC more configurable.   I like having all the config in one file and having the code using the config pull out what it needs.

PR #267, opened 2021-03-10
Closes #258: Added user specified metrics

### LINKED ISSUE #258: Allow user-specified metric to be used with conditional metrics
Currently the metric used in the [conditional metric](https://github.com/mozilla/PRESC/blob/master/presc/evaluations/conditional_metric.py#L111) evaluation is hardcoded. We should allow one or more metrics to be specified by the user. A user-specified metric should include both the function and a display name.

gleonard-m:
Closes #258

Added functionality to allow user to define the metric function used in the analysis.  The metrics used can be specified in the config_default.yaml file 

- Multiple functions with an optional label can be specified.  If no label is provided the function name is used.
```
metrics:
  - function: accuracy_score
    display_name: Accuracy
  - function: jaccard_score
    display_name: Jaccard Score
```
- A per column override is provided.  Specifying per column metrics results in only those functions used as metrics for that column and is not in addition to the default functions.
```
columns:
  density:
    metrics:
      - function: jaccard_score
        display_name: Jaccard Score
```

### DIFF
--- a/presc/config_default.yaml
+++ b/presc/config_default.yaml
@@ -24,8 +24,6 @@ evaluations:
     metrics:
       - function: accuracy_score
         display_name: Accuracy
-      - function: jaccard_score
-        display_name: Jaccard Score
     computation:
       # Number of bins for partitioning a numeric column
       num_bins: 10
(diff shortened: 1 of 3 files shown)


dzeber (approved):
Very nice, clear and concise! Also the test cases do a good job of covering the fallback under failure.

PR #271, opened 2021-04-16
First outline of the copying process.

alberginia:
Testing.


dzeber (approved):
Thanks! Sorry for the delay, I missed this earlier. 

A couple comments, I think we discussed previously:

- Would be nice to colour the points by class in 155
- 238 is a deterministic grid rather than a uniform random sample, but that is one way to do the sampling. Also, you should check out `numpy.meshgrid` and related, might be useful here

PR #272, opened 2021-04-24
SVC copies using blob or moon datasets with three different samplings

alberginia:
1) I generated two problems (with blobs and moons from sklearn) and fitted a linear SVC classifier to each of them (the original classifiers).
2) I assumed the dynamic range of the features is known, so I extracted this information from the original blobs and moon datasets.
3) I generated 6 synthetic datasets: grid-like, random uniform distribution, and random normal distribution, combined with the dynamic range either of the blobs or the moons problems.
4) I used the 6 sets of synthetic data to train 6 linear SVC classifier copies (3 for each problem).
5) I evaluated the performance of the copies with respect to the synthetic data and also with respect to the original data.

I am unclear on how should I scale the covariance in a reasonable way on the normal distributions as a function of the maximum and minimum values of the features, which I am assuming as known.


dzeber:
This is looking good, nice to see the results on the different sampling approaches. I've got a couple of suggestions at this point, and we can discuss more in next week's meeting.

- It would be good to implement the fidelity metric, which looks at the agreement between the original and copy predictions on a dataset (either original or synthetic).
- I think the performance of the copy on the synthetic data is less interesting since it is separable by design. In fact, I wouldn't bother with the train-test split for the synthetic data (you can always generate more).
- Even at this early stage, it's a good idea to reduce code duplication by setting up a function once you have to use the same piece of code multiple times. Then those functions can be moved to the main library.

Seeing the methodology working on these toy problem opens up a whole bunch of further questions for me that we can consider addressing during the course of the project:

- How well does this work for more complex problems?
    - Larger, higher-dimensional, non-separable datasets
    - Non-numerical data
    - More complex models
- How do the metrics vary as a function of synthetic dataset size?
    - I expect this to be most interesting when we get into very high-dimensional datasets
- How do the metrics vary as a function of copy model complexity (parametrized in some approrpiate way)?
- What empirical guidance can we offer on how to select a copy model family for a problem?
- What empirical guidance can we offer on how to select a sampling methodology for a problem?

PR #280, opened 2021-08-18
Add multiple docstrings and tests for the ML copying package

alberginia:
Add docstrings of the sampling functions, the empirical fidelity error, and the copy class.

### DIFF
--- a/presc/copies/copying.py
+++ b/presc/copies/copying.py
@@ -3,6 +3,29 @@ from presc.copies.evaluations import empirical_fidelity_error
 
 
 class ClassifierCopy:
+    """Represents a classifier copy and its associated sampling method of choice.
+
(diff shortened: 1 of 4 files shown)


alberginia:
I have added tests for the functions of the ML classifier copying package.

presc/copies/copying.py
```diff
@@ -3,6 +3,29 @@
 
 
 class ClassifierCopy:
+    """Represents a classifier copy and its associated sampling method of choice.
+
+    Each instance wraps the original ML classifier with a ML classifier copy and
+    the sampling method to carry out the copy. Class methods allow to carry out
```
dzeber, line 9: Super nit: "class method" means something different in Python. Would be best to drop the word "class" here.

dzeber (approved):
Looks good overall. There's one very minor inline comment.

PR #283, opened 2022-02-10
Add PRESC logo to documentation

alberginia:
In this PR I upload to the repository the SVG file of the logo I made for PRESC in its black version with some lateral margin, so that it will not be take up as much space in the menu, and I modify the Sphinx configuration at conf.py in order to include it.

Perhaps it would be useful to add another version of the logo without the margin and/or in its white version, so that it will be available to be used anywhere else. However, I am unsure if this should be the folder to do this.


dzeber:
Not sure about the path issue - it wfm locally. Do you see [this dir](https://github.com/alberginia/PRESC/blob/add-logo/sphinx_docs/source/_images) as a link in your local clone?

alberginia:
Yes I do.

alberginia:
I doesn't appear like the logo is showing up in the documentation, which I don't understand because the link and everything is the same as the one for the scheme, and that image does show up.

dzeber:
As discussed, the documentation build job was being blocked by the CI testing failure. It shows up now :)

PR #287, opened 2022-03-21
Copies with categorical data

alberginia:
This PR corresponds to the introduction of the ability to sample and generate **categorical data** when carrying out copies with the **ML Classifier Copy package**.

In summary I have:
* Added the function `find_categories` to conveniently obtain the categorical features dictionary and its probabilities, when a sample of the original data is available.
* Added the function `build_equal_category_dict` to assign equal probabilities to the categories in the feature dictionary, for the cases when the original data is not available.
* Added the `mixed_data_features` function to build the dictionary with the feature parameters when the original data is available and is a mix of numerical and categorical features.
* Added the `categorical_sampling` function that generates categorical data following a discrete distribution with the probabilities specified in the dictionary of feature parameters.
* Added the function `mixed_data_sampling` that generates datasets of mixed numerical and categorical features according to the dictionary of feature parameters.
* Updated the `labeling` function and the `generate_synthetic_data` method of the `ClassifierCopy` class in order to incorporate the ability to deal with categorical features.
* Added/updated the corresponding docstrings for all the functions.
* Added tests for all the new functions.

### DIFF
--- a/presc/copies/sampling.py
+++ b/presc/copies/sampling.py
@@ -410,7 +410,31 @@ def spherical_balancer_sampling(
 
 
 def categorical_sampling(nsamples=500, random_state=None, feature_parameters=None):
-    """."""
+    """Sample de classifier with a discrete distribution sampling.
+
+    Generates synthetic samples with a discrete distribution according to the
+    probabilities described by `feature_parameters`. Features are assumed to be
+    independent (not correlated).
+
+    Parameters
+    ----------
+    nsamples : int
+        Number of samples to generate.
+    random_state : int
+        Random seed used to generate the sampling data.
+    feature_parameters : dict of dicts
+        A dictionary with an entry per dataset feature (dictionary keys should
+        be the feature names), where each feature entry must contain a
+        nested dictionary with its categories and their fraction. The key for
+        the nested dictionary of categories should be "categories", and the keys
+        for the fractions should be the category name.
+
+    Returns
+    -------
+    pandas DataFrame
+        Dataset with a generated sampling following the discrete distribution of
+        the feature space characterized by the `feature_parameters`.
+    """
     if random_state is not None:
         np.random.seed(seed=random_state)
 
--- a/presc/copies/sampling.py
+++ b/presc/copies/sampling.py
@@ -522,16 +522,10 @@ def mixed_data_sampling(
 
     # Combine the numerical/categorical features respecting the structure of the
     # original dataset
-    X_generated = pd.DataFrame()
-    for feature in feature_parameters:
-        if "categories" in feature_parameters[feature]:
-            X_generated = X_generated.join(
-                X_generated_categorical[[feature]], how="outer"
-            )
-        else:
-            X_generated = X_generated.join(
-                X_generated_numerical[[feature]], how="outer"
-            )
+    X_generated = pd.concat(
+        [X_generated_numerical, X_generated_categorical], axis="columns"
+    )
+    X_generated = X_generated[feature_parameters.keys()]
 
     return X_generated
 
--- a/presc/copies/sampling.py
+++ b/presc/copies/sampling.py
@@ -500,6 +500,7 @@ def mixed_data_sampling(
         else:
             features_numerical.append(feature)
 
+    # Generate feature parameter dictionaries for the numerical/categorical samplers
     feature_parameters_numerical = {
         feature: feature_parameters[feature] for feature in features_numerical
     }
@@ -507,6 +508,7 @@ def mixed_data_sampling(
         feature: feature_parameters[feature] for feature in features_categorical
     }
 
+    # Generate the numerical/categorical features of each sample separately
     X_generated_numerical = numerical_sampling(
         nsamples=nsamples,
         random_state=random_state,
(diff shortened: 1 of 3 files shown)


presc/copies/sampling.py
```diff
@@ -40,6 +41,118 @@ def dynamical_range(df):
     return range_dict
 
 
+def find_categories(df, add_NaNs=False):
```
dzeber, line 44: This is a nit, but I'd prefer simple lower-casing `add_nans`. Stylistically it's more uniform and reduces potential for confusion/bugs.

presc/copies/sampling.py
```diff
@@ -40,6 +41,118 @@ def dynamical_range(df):
...
+        column names), where each feature entry contains a nested dictionary
+        with its categories and the fraction of each category present in the
+        analyzed dataset (the nested dictionary key for this information is
+        "categories", which is also a dictionary with one entry per category).
+    """
+    categories_dict = {}
+    for feature in df:
+        if is_discrete(df[feature]):
+            # Remove NaN values from selection
+            df_no_nans = df[df[feature].notnull()]
```
dzeber, line 68: I see you're using `value_counts` inside the for loop below. In fact, all of this computation could be handled by a single call to `value_counts`. Eg.
```python
value_props = df[feature].value_counts(normalize=True, dropna=(not add_nans))
```
Then `value_props` is Series with an entry for each unique value listing its relative frequency, and you can do
```python
categories_dict[feature] = {
    "categories": value_props.to_dict()
}
```
It'll handle the cases around NaNs implicitly using `dropna`.

presc/copies/sampling.py
```diff
@@ -296,6 +409,133 @@ def spherical_balancer_sampling(
     return df_generated
 
 
+def categorical_sampling(nsamples=500, random_state=None, feature_parameters=None):
```
dzeber, line 412: (here and below) if `feature_parameters` is actually required, it would be better to put it first in the param list with no default value, to make that clear.

presc/copies/sampling.py
```diff
@@ -296,6 +409,133 @@ def spherical_balancer_sampling(
...
+        feature_parameters=feature_parameters_categorical,
+    )
+
+    # Combine the numerical/categorical features respecting the structure of the
+    # original dataset
+    X_generated = pd.DataFrame()
+    for feature in feature_parameters:
+        if "categories" in feature_parameters[feature]:
+            X_generated = X_generated.join(
+                X_generated_categorical[[feature]], how="outer"
```
dzeber, line 529: Is there a reason to join each column separately? Since you've already got the two `X_generated` DFs, is this different from concatenating them together? Eg.
```python
pd.concat([X_generated_numerical, X_generated_categorical], axis="columns")
```

dzeber (approved):
This looks fine and covers the categorical sampling use case. I've left a few comments about implementation/stylistic details inline.

presc/copies/sampling.py
alberginia: The reason is to maintain the same order of the original columns. With a simple block concatenation the categorical and the numerical columns become separated. I don't know a priori if the model will use the names of the columns (in which case the order wouldn't matter) or if the model will take the same order of the parameters in which it was trained, so I wanted to make sure that the structure of the sampled dataset was exactly the same.

presc/copies/sampling.py
alberginia: Done.

presc/copies/sampling.py
alberginia: I changed this but then I realised that this also affects the `ClassifierCopy` class, which has a `**k_sampling_parameters` that includes the `feature_parameters` keyword argument, among others, and I am unsure how to proceed if this particular argument is then positional.

Should I add another argument (e.g. `*sampling_parameters`) to allow for the positional `feature_parameters` argument? Where exactly should I place this?

Maybe if this complicates things too much I could simply leave `feature_parameters` as a keyword argument, but add a note/error that the keyword argument is required?  :thinking:

presc/copies/sampling.py
alberginia: I leave it as it is for the moment, as we agreed in our discussion.

presc/copies/sampling.py
alberginia: @dzeber, I tried to change the `feature_parameters` argument in all the sampling functions and leaving the `ClassifierCopy` class unchanged, as you suggested, but I get an error in the two tests involving the class that says "got multiple values for argument 'feature_parameters'" so there is something wrong.

dzeber: Looks like you still need to update where `mixed_data_sampling` is getting called.

presc/copies/sampling.py
dzeber: As discussed, you can reorder the columns of a DF by passing a list of column names in the desired order, eg. `X[["col2", "col1"]]`.

presc/copies/sampling.py
alberginia: I've now finally understood and managed to fix what was wrong with the `mixed_data_sampling` function call within the `ClassifierCopy` class.  :)

As far as I can see, the new code is passing the tests of the ML Classifier Copying package. I think the CI fails for some other reason.

presc/copies/sampling.py
alberginia: Done.

PR #292, opened 2022-09-22
Best practices document

alberginia:
This PR corresponds to Milestone 11 of the PRESC ML Classifier Copies Package. 

It incorporates a document with Best Practices and Recommendations to the users on the best strategies to carry out the classifier copying process for each case, depending on the datasets and the classifiers. 

This document summarizes the practical know-how of carrying out a numerous diversity of ML classifier copies in different contexts and with diverse problems.

The content of the document has been simply added together as a section of the package documentation.

**TASKS:**

Develop a document explaining best practices and recommendations to the users:
* Best practices document


docs/ml_copies.md
```diff
@@ -70,3 +70,70 @@ The Replacement Capability is the ratio between the accuracy of the copy model w
...
+ 
+Although ML classifier copies can be carried out without any access to the original data, at least some assumptions about the relevant range of the features is necessary. The more precise the assumptions that can be made, the more efficient the sampling will be, which is specially relevant in problems with a large number of dimensions.
+
+If the feature space descriptors are extracted from original data, be careful with outliers because they can enormously and unnecessarily increase the feature space that will need to be sampled. It is a good practice to always review any feature space descriptors that were automatically extracted from the original data, in order to reduce excessively wide ranges and to correct obvious misalignments.
+
+
+ * **Choice of sampler**
+ 
+Proper sampling of the feature space is one of the key issues to ensure a successful copy, but it quickly becomes very challenging as the number of features increase. If we have no information on the feature distributions it is a good idea to use a uniform sampler. but this may quickly become very unefficient as the number of features increase, where a normal or a spherical sampler may be more appropriate.
```
dzeber, line 117: "unefficient" -> "inefficient"

docs/ml_copies.md
```diff
@@ -70,3 +70,70 @@ The Replacement Capability is the ratio between the accuracy of the copy model w
...
+
+### Sampling
+
+ * **Standardize the features**
+
+Certain samplers, such as the spherical sampler in this package, may not handle well a large difference in the absolute magnitude of the different features, or that they are not centered in the origin. Therefore, it is advisable to standardize the input data in case of doubt. However, it is unavoidable with this sampler, that classes in more centric regions will be better sampled that more peripherical ones.
+
+ * **Reduce the features space**
+ 
+Although ML classifier copies can be carried out without any access to the original data, at least some assumptions about the relevant range of the features is necessary. The more precise the assumptions that can be made, the more efficient the sampling will be, which is specially relevant in problems with a large number of dimensions.
```
dzeber, line 110: I would emphasize here that the data that we're generating through sampling needs to be accepted as input for prediction by the original model, so we need to something about the schema it expects, eg. data types, ranges, etc.

docs/ml_copies.md
```diff
@@ -70,3 +70,70 @@ The Replacement Capability is the ratio between the accuracy of the copy model w
...
+ * **Handle imbalance**
+ 
+In ML classifier copies, imbalance between classes arises when they are sampled inequally, either because they occupy very disparate volumes of the feature space, as in the case of uniform samplers, or because the sampler does not explore the feature space homogeneously. Hence, it is also a good practice to ensure that a similar number of synthetic samples are generated for all classes. 
+
+
+### Copy model characteristics
+
+ * **Choice of copy model family**
+
+Every model family is able to shape the decision boundary in a different manner. If the goal is to carry out an exact copy of the original model with the lowest empirical fidelity error, it must be ensured that the copy model family has the ability to mimic the shape of the decision boundary in the same manner as the original model. For instance, decision tree classifiers describe decision boundaries with horizontal and vertical edges, but have a harder time to perfectly mimic models with slanted edges and viceversa, other model families may have a hard time mimicking such abrupt angles.
```
dzeber, line 128: "horizontal and vertical" -> "piecewise constant boundaries"
"slanted" -> "general linear boundaries"

docs/ml_copies.md
```diff
@@ -70,3 +70,70 @@ The Replacement Capability is the ratio between the accuracy of the copy model w
...
+       
+Conversely, if the goal of the copy is mainly to be able to replace the original model with a classifier of a similar or better performance, then the copy model family must have the ability to describe the inherent topology of the problem instead. In some cases, if the original model was not the best choice to describe the problem's inherent decision boundary, a good choice of the copy model family may even surpass the original in accuracy and yield a replacement capability larger than one.
+
+
+ * **Complexity of the copy model**
+
+The ML classifier copy problem is always separable, and overfitting is not an issue. An arbitrarily large number of samples can in principle be generated in order to ensure an appropriate coverage of the feature space to properly fit the copy classifier. However, the complexity capability of the copy classifier must be at least as large or more than the original classifier to obtain a good copy. 
+       
+When the performance of the copy gets stuck and falls short of the original classifier, despite an increasing number of generated samples, it is an indication that the complexity of the copy model may be too low.
+
```
dzeber, line 138: Want to have a comment about what happens is the complexity is set too high?

docs/ml_copies.md
```diff
@@ -70,3 +70,70 @@ The Replacement Capability is the ratio between the accuracy of the copy model w
...
+       
+Conversely, if the goal of the copy is mainly to be able to replace the original model with a classifier of a similar or better performance, then the copy model family must have the ability to describe the inherent topology of the problem instead. In some cases, if the original model was not the best choice to describe the problem's inherent decision boundary, a good choice of the copy model family may even surpass the original in accuracy and yield a replacement capability larger than one.
+
+
+ * **Complexity of the copy model**
+
+The ML classifier copy problem is always separable, and overfitting is not an issue. An arbitrarily large number of samples can in principle be generated in order to ensure an appropriate coverage of the feature space to properly fit the copy classifier. However, the complexity capability of the copy classifier must be at least as large or more than the original classifier to obtain a good copy. 
+
```
dzeber, line 136: Can we make a recommendation for how to select a copy model class when we don't know much about the original model?

dzeber (commented):
I like this - I think you've covered the main relevant points. I've got some suggestions for rephrasing and places to expand slightly.

docs/ml_copies.md
alberginia: I don't think so. It is highly dependent on the problem. I'd say you need to try and check on the performance on the copy to figure out what to do next.

docs/ml_copies.md
```diff
@@ -70,3 +70,70 @@ The Replacement Capability is the ratio between the accuracy of the copy model w
...
+Do you only want to ensure you get a model with the performance as good or better as the original model but maintaining consistency is not important? In this case, use the copy replacement capability to evaluate your copy.
+
+
+### Model selection
+
+ * **Test data**
+ 
+If some labeled original test data is available, it is always more reliable to use that to compute the metrics and carry out the copy evaluation. Even if that original data is not labeled, it can always be labeled using the prediction of the original model and then be used to evaluate the copy. But this will at least ensure that the sampling is done following the original feature distribution.
+       
+If no original data is available, then a test synthetic dataset can be generated for evaluation of the copy. But be aware that, if this dataset is generated with the same sampler that was used to carry out the copy, the test will focus on the same regions where the copy has had more training. Hence, the fidelity or the performance of the copy will not be evaluated as rigorously on the outskirt decision boundary.
```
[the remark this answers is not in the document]
alberginia: 1) Fixed.
2) I've tried to clarify the sentence.

docs/ml_copies.md
```diff
@@ -70,3 +70,70 @@ The Replacement Capability is the ratio between the accuracy of the copy model w
...
+
+ * **Sampling in evaluation**
+
+What does it mean if the copy has a low empirical fidelity error with synthetic data, but at the same time has a high fidelity error with test data? 
+
+The copy is often created without having a priori an idea of the distribution of the real data, this means that real data may be located mostly in a particular region while synthetic data will tend to be all over the place. This means that the original model which has been trained with real data will have a stronger emphasis on having certain boundaries right, while the copy model will be trained with data that enforces to have a similar boundary as the original mostly everywhere.
```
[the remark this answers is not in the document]
alberginia: I don't think it is necessarily a higher resolution. This would happen only if data is covering more parts of the feature space in one case than in the other. What I meant was that a higher number of samples in a certain region will force the algorithm to be "right" there (because of a higher weight in the loss function) and perhaps be more flexible about being wrong elsewhere.

docs/ml_copies.md
alberginia: Corrected.

docs/ml_copies.md
alberginia: Corrected.

docs/ml_copies.md
alberginia: Theoretically, complexity isn't a problem in the sense that we can get as many samples as we want. In practice it may a different matter because it could make the copy process more inefficient.

docs/ml_copies.md
alberginia: Added.

docs/ml_copies.md
alberginia: I'm don't think this is particularly problematic, other than possibly taking a while longer and having to generate more samples, because we cannot overfit copies.

docs/ml_copies.md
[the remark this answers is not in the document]
dzeber: Agreed, that's what I meant by "higher resolution" :)

PR #296, opened 2022-10-30
ML Classifier Copies Autoencoder

alberginia:
This PR corresponds to Milestone 13 of the PRESC ML Classifier Copies Package. A solution has been developed for sampling the feature space to perform the copy when there are a large number of dimensions but a much smaller manifold. This is a common problem but is particularly extreme in the case of image problems. How to sample the relevant feature space when there are a lot of dimensions is one of the main problems in ML classifier copying.

An example of a convolutional variational autoencoder has been implemented here in order to help carry out ML classifier copies of image problems.

The new `autoencoders` module includes:
* **class VAE**: General Variational Autoencoder model built from the encoder and decoder models, with a custom train step and metrics. Namely, a reconstruction loss and a Kullback-Leibler divergence loss, that together account for the total loss. 
* **class Sampling**: Layer implementing the Gaussian distribution sampling of the Variational Autoencoder in the latent space.
* **class ImageVAE**: Implements a Convolutional Variational Autoencoder class for image samples.
* **class Orthonormal**: Constraint class to approximate orthonormal weights during autoencoder training.
* **plot_autoencoder_latent_space function**: Function to display two dimensions of the image representation of the autoencoder in the latent space.

Also, two image sampling functions were added to the `sampling` module:
* **image_random_sampling**: Generates random Gaussian noise image samples. 
* **image_vae_sampling**: Generates random samples from the Gaussian distributions obtained by an image autoencoder in the latent space representation.

Complete docstrings for all the functions and classes were added as well.


dzeber:
This looks good. We should tensorflow to the dev environment - I'll update that.

PR #297, opened 2022-10-30
ML Classifier Copies VAE sampler example

alberginia:
This PR corresponds to Milestone 14 of the PRESC ML Classifier Copies Package.

A Jupyter notebook example that applies the solution implemented in the autoencoder module to an example using a "black box" classifier model of the MNIST Handwritten Digits dataset to show the users how a variational autoencoder sampler can be used in difficult copy problems with many dimensions.

Another dataset is used to train the autoencoder that then can better explore the latent feature space to carry out the copy than blind random sampling.


dzeber:
Nice work on the notebook. I think this gives a nice demonstration of the power of the approach using the autoencoder. One recommendation for the future would be to rework the DigitsModel data loading to load from tf.keras.datasets rather than download the full raw files directly, now that TF is in the environment.

